PR #13, opened 2018-01-01
Proof of concept: async/await

eirslett:
A lot of the code in this library could be simplified by using the async/await syntax instead of nesting Promise `.then()` calls. I only did a couple of functions, as a proof of concept. If this is something you would like to do?

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,7 @@
+root = true
+
+[*]
+end_of_line = lf
+indent_style = tab
+insert_final_newline = true
+trim_trailing_whitespace = true
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -17,7 +17,7 @@ module.exports =
 	settings: {},
 	parserOptions:
 	{
-		ecmaVersion: 6,
+		ecmaVersion: 8,
 		sourceType: 'module',
 		ecmaFeatures:
 		{
--- a/lib/Transport.js
+++ b/lib/Transport.js
@@ -181,7 +181,7 @@ export default class Transport extends EnhancedEventEmitter
 		this._handler.close();
 	}
 
-	restartIce()
+	async restartIce()
 	{
 		logger.debug('restartIce()');
 
@@ -190,27 +190,20 @@ export default class Transport extends EnhancedEventEmitter
 		else if (this._connectionState === 'new')
 			return;
 
-		Promise.resolve()
-			.then(() =>
+		try
+		{
+			const data =
 			{
-				const data =
-				{
-					id : this._id
-				};
+				id : this._id
+			};
+			const { remoteIceParameters } = await this.safeEmitAsPromise('@request', 'restartTransport', data);
 
-				return this.safeEmitAsPromise('@request', 'restartTransport', data);
-			})
-			.then((response) =>
-			{
-				const remoteIceParameters = response.iceParameters;
-
-				// Enqueue command.
-				return this._commandQueue.push('restartIce', { remoteIceParameters });
-			})
-			.catch((error) =>
-			{
-				logger.error('restartIce() | failed: %o', error);
-			});
+			return this._commandQueue.push('restartIce', { remoteIceParameters });
+		}
+		catch (error)
+		{
+			logger.error('restartIce() | failed: %o', error);
+		}
 	}
 
 	enableStats(interval = DEFAULT_STATS_INTERVAL)
@@ -620,38 +613,25 @@ export default class Transport extends EnhancedEventEmitter
 		promiseHolder.promise = promise;
 	}
 
-	_execAddProducer(producer)
+	async _execAddProducer(producer)
 	{
 		logger.debug('_execAddProducer()');
 
-		let producerRtpParameters;
-
 		// Call the handler.
-		return Promise.resolve()
-			.then(() =>
-			{
-				return this._handler.addProducer(producer);
-			})
-			.then((rtpParameters) =>
-			{
-				producerRtpParameters = rtpParameters;
+		const producerRtpParameters = await this._handler.addProducer(producer);
 
-				const data =
-				{
-					id            : producer.id,
-					kind          : producer.kind,
-					transportId   : this._id,
-					rtpParameters : rtpParameters,
-					paused        : producer.locallyPaused,
-					appData       : producer.appData
-				};
+		const data =
+		{
+			id            : producer.id,
+			kind          : producer.kind,
+			transportId   : this._id,
+			rtpParameters : producerRtpParameters,
+			paused        : producer.locallyPaused,
+			appData       : producer.appData
+		};
 
-				return this.safeEmitAsPromise('@request', 'createProducer', data);
-			})
-			.then(() =>
-			{
-				producer.setRtpParameters(producerRtpParameters);
-			});
+		await this.safeEmitAsPromise('@request', 'createProducer', data);
+		producer.setRtpParameters(producerRtpParameters);
 	}
 
 	_execRemoveProducer(producer)
@@ -670,47 +650,33 @@ export default class Transport extends EnhancedEventEmitter
 		return this._handler.replaceProducerTrack(producer, track);
 	}
 
-	_execAddConsumer(consumer)
+	async _execAddConsumer(consumer)
 	{
 		logger.debug('_execAddConsumer()');
 
-		let consumerTrack;
-
 		// Call the handler.
-		return Promise.resolve()
-			.then(() =>
-			{
-				return this._handler.addConsumer(consumer);
-			})
-			.then((track) =>
-			{
-				consumerTrack = track;
+		const consumerTrack = await this._handler.addConsumer(consumer);
 
-				const data =
-				{
-					id               : consumer.id,
-					transportId      : this.id,
-					paused           : consumer.locallyPaused,
-					preferredProfile : consumer.preferredProfile
-				};
+		const data =
+		{
+			id               : consumer.id,
+			transportId      : this.id,
+			paused           : consumer.locallyPaused,
+			preferredProfile : consumer.preferredProfile
+		};
 
-				return this.safeEmitAsPromise('@request', 'enableConsumer', data);
-			})
-			.then((response) =>
-			{
-				const { paused, preferredProfile, effectiveProfile } = response;
+		const { paused, preferredProfile, effectiveProfile } = await this.safeEmitAsPromise('@request', 'enableConsumer', data);
 
-				if (paused)
-					consumer.remotePause();
+		if (paused)
+			consumer.remotePause();
 
-				if (preferredProfile)
-					consumer.remoteSetPreferredProfile(preferredProfile);
+		if (preferredProfile)
+			consumer.remoteSetPreferredProfile(preferredProfile);
 
-				if (effectiveProfile)
-					consumer.remoteEffectiveProfileChanged(effectiveProfile);
+		if (effectiveProfile)
+			consumer.remoteEffectiveProfileChanged(effectiveProfile);
 
-				return consumerTrack;
-			});
+		return consumerTrack;
 	}
 
 	_execRemoveConsumer(consumer)


.editorconfig
```diff
@@ -0,0 +1,7 @@
+root = true
+
+[*]
+end_of_line = lf
+indent_style = tab
```
eirslett: A bit unrelated maybe, but I added a .editorconfig that helps enforce eslint config - especially tab indentation.

.eslintrc.js
```diff
@@ -17,7 +17,7 @@ module.exports =
 	settings: {},
 	parserOptions:
 	{
-		ecmaVersion: 6,
+		ecmaVersion: 8,
```
eirslett: await/async is from a new JS standard.

ibc:
Definitely I must learn about `async`/`await`, but I prefer to defer it for a bit later (not before 2019!). So thanks, let's hold this PR open as a reminder.

thehappycoder:
async / await is not that hard, so there is no reason to hold it, I think

ibc:
Well, I've limited time and priorities other than learning async/await right now. I will do when possible.

Sharcoux:
I don't think that async/await is better than the use of Promise. The issue here is more that the use of Promise is not done correctly and could be improved. `Promise.resolve().then(...)` for instance is totally unnecessary. Using `let myVar` outside the promise and then giving it a value inside the promise is not a good pattern. You can simply set getMyVar to the value of your promise and access it with `getMyVar().then(value => xxx)`. And finally, mixing promises and try/catch blocks is not a good idea as `promise.then` and `promise.catch` already accomplish this.

ibc:
> The issue here is more that the use of Promise is not done correctly and could be improved.

Do you mean in the current mediasoup-client code? If so, can you point where it's not correct?


> Promise.resolve().then(...) for instance is totally unnecessary.

I know. It's just cosmetic, but I don't think it's "incorrect".


> Using `let myVar` outside the promise and then giving it a value inside the promise is not a good pattern. You can simply set `getMyVar` to the value of your promise and access it with `getMyVar().then(value => xxx)`.

Can you provide an example using that pattern? I mean, take some code of mediasoup-client code and convert it to that pattern.

> And finally, mixing promises and try/catch blocks is not a good idea as `promise.then` and `promise.catch` already accomplish this.

Where does that happen in the code?

eirslett:
The code works today, even if there are some strange uses of Promise... If there’s enough time to refactor anyways, I suggest going for async/await (which is just a much cleaner syntax for Promises, it’s still Promises) since that’s generally considered a more future-oriented pattern. Just my 2c, but I’m not the one doing the hard work here, so don’t value my opinion too much.

Sharcoux:
@ibc I'm trying to make a PR or sth to show you what I mean, but anyway, as said @eirslett the code works as it is. My comment is just aiming at defending Promises towards async await, because I find it much clearer when used correctly, and I don't criticize the work made by @ibc but only pointing out some general improvement in the code style.

eirslett:
> I find it much clearer when used correctly, and I don't criticize the work made by @ibc but only pointing out some general improvement in the code style.

@Sharcoux I'm curious about what you mean. How would you express the code below with pure promises, in a cleaner way?

```javascript
async function doStuff () {
  const a = await getA();
  const b = await getB(a);
  const c = await getC(a, b);
  const d = await getD(a, b, c);
  return a + b + c + d;
}
```

(Sorry for derailing the topic somewhat)

Sharcoux:
Something like that, with better formatting
`return getA().then(a => getB(a).then(b => getC(a,b).then(c => getD(a,b,c).then(d => a+b+c+d))))`

Sharcoux:
this is how I would write _execAddProducer:

```
  _execAddProducer(producer) {
    logger.debug('_execAddProducer()');

    // Call the handler.
    return this._handler.addProducer(producer).then(rtpParameters => {
      const data = {
        id: producer.id,
        kind: producer.kind,
        transportId: this._id,
        rtpParameters: rtpParameters,
        paused: producer.locallyPaused,
        appData: producer.appData,
      };

      return this.safeEmitAsPromise('@request', 'createProducer', data).then(
        () => producer.setRtpParameters(rtpParameters)
      );
    });
  }

```

instead of current

	_execAddProducer(producer)
	{
		logger.debug('_execAddProducer()');

		let producerRtpParameters;

		// Call the handler.
		return Promise.resolve()
			.then(() =>
			{
				return this._handler.addProducer(producer);
			})
			.then((rtpParameters) =>
			{
				producerRtpParameters = rtpParameters;

				const data =
				{
					id            : producer.id,
					kind          : producer.kind,
					transportId   : this._id,
					rtpParameters : rtpParameters,
					paused        : producer.locallyPaused,
					appData       : producer.appData
				};

				return this.safeEmitAsPromise('@request', 'createProducer', data);
			})
			.then(() =>
			{
				producer.setRtpParameters(producerRtpParameters);
			});
	}

eirslett:
```
async function doStuff () {
  return Promise.all([getA, getB, getC, getD]).then(result => result.reduce((sum, x) => sum + x));
}
```
This wouldn't work, because getB needs the result from getA and so on, so they cannot run in parallel... There are some requirements like that in mediasoup as well, functions cannot be called in parallel because one function depends on the result of another. Using only Promises, you end up with [callback hell](http://callbackhell.com/)...

ibc:
> Using only Promises, you end up with callback hell...

IMHO using Promises chain removes the "callback hell". I don't mean that it looks better than async/await, but when I see a promise chain I clearly understand the code.

ibc:
### Current style

```js
_execAddProducer(producer)
{
	logger.debug('_execAddProducer()');

	let producerRtpParameters;

	// Call the handler.
	return Promise.resolve()
		.then(() =>
		{
			return this._handler.addProducer(producer);
		})
		.then((rtpParameters) =>
		{
			producerRtpParameters = rtpParameters;

			const data =
			{
				id            : producer.id,
				kind          : producer.kind,
				transportId   : this._id,
				rtpParameters : rtpParameters,
				paused        : producer.locallyPaused,
				appData       : producer.appData
			};

			return this.safeEmitAsPromise('@request', 'createProducer', data);
		})
		.then(() =>
		{
			producer.setRtpParameters(producerRtpParameters);
		});
}
```

### @Sharcoux style

```js
_execAddProducer(producer) 
{
	logger.debug('_execAddProducer()');

	// Call the handler.
	return this._handler.addProducer(producer)
		.then((rtpParameters) =>
		{
			const data =
			{
				id            : producer.id,
				kind          : producer.kind,
				transportId   : this._id,
				rtpParameters : rtpParameters,
				paused        : producer.locallyPaused,
				appData       : producer.appData,
			};

			return this.safeEmitAsPromise('@request', 'createProducer', data)
				.then(() => producer.setRtpParameters(rtpParameters));
		});
}
```

Yes, the current style takes more lines of code. But if you remove the initial and cosmetic `return Promise.resolve()` then they both are similar.

Regarding the bottom style, honestly I don't like promises resolution within a `then()` body. Yes, my approach requires having `let producerRtpParameters;` declared at the top of the full scope but, is it so hard?

Sharcoux:
Actually, going a little further, I would improve the execCommand function like that:

```javascript

  _execCommand(command, callback) {
    //get the corresponding function
    const functionImpl = this['_' + command.method + 'Impl'];
    //call the callback function with the resulting promise
    callback(
      functionImpl
        ? functionImpl(command)
        : Promise.reject(
            new Error(`unknown command method "${command.method}"`)
          )
    );
  }

  _addProducerImpl({ producer }) {...}
  _removeProducerImpl({ producer }) {...}
  _replaceProducerTrackImpl({producer, track}) {...}
  ...
```

And in CommandQueue, you would need to make these changes:

```javascript
  _handleCommand(command) {
    logger.debug('_handleCommand() [method:%s]', command.method);

    if (this._closed) command.reject(new InvalidStateError('closed'));

    const callback = promise =>
      promise
        .then(result => {
          logger.debug(
            '_handleCommand() | command succeeded [method:%s]',
            command.method
          );
          if (this._closed) command.reject(new InvalidStateError('closed'));
          else command.resolve(result);
        })
        .catch(error => {
          logger.error(
            '_handleCommand() | command failed [method:%s]: %o',
            command.method,
            error
          );

          // Reject the command with the error.
          command.reject(error);
        });
    this.emit('exec', command, callback);
  }
```

Sharcoux:
@ibc Once again, your code is good, it works, and the point of the discussion, here, is only cosmetic and patterns, and I started talking about this only because the PullRequest is about cosmetic changes.

I am sorry for using 'incorrect' which was not the right term. Your approach is correct. I just think it is not optimal and that is why @eirslett is proposing the use of async await whereas the use of promise would look almost the same if you get rid of the unnecessary parts, and that was the only motive of my initial comment.

It is your project so it is totally up to you if you want to write:
```javascript
return Promise.resolve()
		.then(() =>
		{
			return this._handler.addProducer(producer);
		})
```
But I admit that I don't really see the point though, even for cosmetic.

About `let producerRtpParameters;`, I agree that it is 'not so hard', and it is totally fine. Personally though, I think that it is better practice to avoid extending the scope of variables out of what is strictly necessary.

Once again, all of this is only subjective and doesn't aim at being a criticism of your great work. Sorry if it looked like that.

Sharcoux:
@eirslett What I was trying to say is that normally you should have sth like that:

```javascript
function doStuff () {
  return Promise.all([getA(), getB(), getC(), getD()]).then(result => result.reduce((sum, x) => sum + x));
}

let a = null;
function getA() {
    return a || (a = doAsyncStuffToGetA());
}
function doAsyncStuffToGetA() {
    return new Promise(resolve => {
        setTimeout(() => resolve('a'), 2000);
    });
}

let b = null;
function getB() {
    return b || (b = doAsyncStuffToGetB());
}
function doAsyncStuffToGetB() {
    return getA().then(a => synchronizedCode(a));
}

let c = null;
function getC() {
    return c || (c = doAsyncStuffToGetC());
}
function doAsyncStuffToGetC() {
    return Promise.all([getA, getB]).then([a,b] => synchronizedCode(a,b));
}

let d = null;
function getD() {
    return d || (d = doAsyncStuffToGetD());
}
function doAsyncStuffToGetD() {
    return Promise.all([getA, getB, getC]).then([a,b,c] => synchronizedCode(a,b,c));
}
```
And here, as you see, no callback hell ;)

eirslett:
Browser support: https://caniuse.com/#feat=async-functions
You can ship async/await untranspiled to browsers, unless you need to support IE 11. But if needed, babel will transpile it to promises.

ibc:
Hi, I've experimented a bit with async/await. It's really amazing as it avoids having to declare empty `let variable;` before promises chain when you need to write such a variable once a promise has resolved.

In fact, I'm using async/await in the [test units](https://github.com/versatica/mediasoup-client/blob/v3/__test__/test.js?ts=2) of mediasoup-client v3: 

Said that:

* Current mediasoup-client v2 is ES6 code babelized to ES5 under `/lib-es5` folder.
* "main" in `package.json` points to `lib-es5/index.js` as entry point.
* I don't like it since the new way to go in web development is about producing libs that expose pure ES6 and it's up to the app to babelize them if needed.
   * Also, some related side issues: #60

So, in mediasoup-client v3:

* No babel stuff is included.
* No `lib-es5` folder is generated (this fixes #43) so no ES5 code is generated.
* It must work in Node so no ES6 `import`/ `export` is used. Instead, CommonJS `require()` is used.
* It must work in Node >= 6.4, but Node 6 does **NOT** support async/await, so mediasoup-client is internally built with promises (as v2).

Of course, the app can use async/await stuff with mediasoup-client v3 (as shown in the [test units](https://github.com/versatica/mediasoup-client/blob/v3/__test__/test.js?ts=2)).

Once we drop support for Node 6, we can migrate some internal code to async/await, but for now better cover as much Node versions as possible.

Sharcoux:
Just a remark: using `let variable;` is not the correct pattern with Promise anyway...

ibc:
> Just a remark: using `let variable;` is not the correct pattern with Promise anyway...

How would you improve [this](https://github.com/versatica/mediasoup-client/blob/v3/lib/handlers/Chrome70.js#L133) (the `let transceiver` and `let localSdpObject`) with just promises? Note they get a value within some promise bodies and then are needed by next promises.

Sharcoux:
about `let transceiver`, the code to access it is synchronous, so there is no reason to delay it's affectation. About `localSdpObject`, check your answer comparing "curent style" and "Sharcoux style". I would solve it the same way that I solved it for `producerRtpParameters`

Sharcoux:
About `transceiver`, just replace `let transceiver` by the following.

```
const transceiver = this._pc.getTransceivers().find((t) => t.receiver.track.kind === track.kind && t.direction === 'inactive');
```
I see no valid reason to delay the affectation.

> Well, let me say that I do not like promises within promises :)

This is really just a question of presentation and architecture and unrelated to promise VS async/await

Would you rather this ?

```
_execAddProducer(producer) {
	return this._handler.addProducer(producer)
		.then(rtpParameters => this.createProducer(producer, rtpParameters))
                 .then(() => producer.setRtpParameters(rtpParameters));
}

createProducer(producer, rtpParameters) {
	const data = {
		id            : producer.id,
		kind          : producer.kind,
		transportId   : this._id,
		rtpParameters : rtpParameters,
		paused        : producer.locallyPaused,
		appData       : producer.appData,
	};

	return this.safeEmitAsPromise('@request', 'createProducer', data);
}
```

You would have the same issue with async/await.

ibc:
Did you see 96a49d5d927dbc736ebcfbc7c39f6855a10fbbbf, @eirslett and @Sharcoux? :)

eirslett:
Congratulations! How do you like it?

ibc:
Definitely a lot. It happens that:

* Browser implementing Promises do also implement async/await.
* Node stable (8.x.x) supports Promises and async/await.

So there are no excuses to keep using Promise chains :)

PR #17, opened 2018-02-04
RN and simulcast

ivmos:
I've detected 2 issues playing with react-native-webrtc and simulcast, first one is missing RTCSessionDescription wrapping in setLocalDescription. I'm not so sure about the second one so let me paste monkeypatch and let me know what you think:

```
diff --git a/lib/Producer.js b/lib/Producer.js
index b0a5d70..081e039 100644
--- a/lib/Producer.js
+++ b/lib/Producer.js
@@ -64,7 +64,7 @@ export default class Producer extends EnhancedEventEmitter
                this._simulcast = false;

                if (options.simulcast)
-                       this._simulcast = Object.assign({}, SIMULCAST_DEFAULT, options.simulcast);
+                       this._simulcast = Object.assign({}, SIMULCAST_DEFAULT, options);

                // Associated Transport.
                // @type {Transport}
```

Without this change I'm getting "In this environment the sources for assign MUST be an object. This error is a performance optimization and not spec compliant." in that Object.assign statement.

Thanks in advance

### DIFF
--- a/lib/handlers/ReactNative.js
+++ b/lib/handlers/ReactNative.js
@@ -137,7 +137,9 @@ class SendHandler extends Handler
 					'addProducer() | calling pc.setLocalDescription() [offer:%o]',
 					offer);
 
-				return this._pc.setLocalDescription(offer);
+				const offerDesc = new RTCSessionDescription(offer);
+
+				return this._pc.setLocalDescription(offerDesc);
 			})
 			.then(() =>
 			{
@@ -303,7 +305,9 @@ class SendHandler extends Handler
 					'replaceProducerTrack() | calling pc.setLocalDescription() [offer:%o]',
 					offer);
 
-				return this._pc.setLocalDescription(offer);
+				const offerDesc = new RTCSessionDescription(offer);
+
+				return this._pc.setLocalDescription(offerDesc);
 			})
 			.then(() =>
 			{


lib/handlers/ReactNative.js
```diff
@@ -137,7 +137,9 @@ class SendHandler extends Handler
 					'addProducer() | calling pc.setLocalDescription() [offer:%o]',
 					offer);
 
-				return this._pc.setLocalDescription(offer);
+				const offerDesc = new RTCSessionDescription(offer);
```
ibc: I will move this to the `if` block above since, if simulcast is not used, `offer` is already a `RTCSessionDescription` object.

ibc: No, it's ok as you did.

ibc:
Done, thanks!

Regarding the `Object.assign()` issue, I need to understand where the problem is. I don't understand this change:

```
-                       this._simulcast = Object.assign({}, SIMULCAST_DEFAULT, options.simulcast);
+                       this._simulcast = Object.assign({}, SIMULCAST_DEFAULT, options);
```

`this._simulcast` must be filled with `options.simulcast` which, as per the [documentation](https://mediasoup.org/documentation/mediasoup-client/api/#Producer-ProducerOptions), must be an object.

So, if `options.simulcast` is an Object, there shouldn't be any error in the code above, do I miss something?

ibc:
How does the `options.simulcast` you give to the Producer in RN?

ivmos:
@ibc Thanks! I'm checking deeper what happens with options.simulcast

ibc:
The problem is that the app is passing true as value of options.simulcast, which works in Chrome but fails in RN. I'll fix it.

ivmos:
Ok thanks!

ibc:
Fixed in 2.0.10: e01aea4aae6b6fb721571b3f533a1f8a11f36186

PR #29, opened 2018-05-07
Add a simple function to update Ice Servers / Turn Server

misi:


ibc:
>```js
>this._settings.turnServers=turnServers;
>```

Hi, obviously by just changing a value in an Object things are not gonna work automatically. This requires updating the corresponding `RTCPeerConnection` instances (or `RTCIceTransport` instances in EDGE).

I'll work on this once I get some spare time.

misi:
Sorry,
I was confused because I didn't read the code carefully. My fault sorry.
My assumption was that restartTransport replace PC, but now I see that it restarts only ICE.

ibc:
Yep, it would require calling `peerconnection.updateSettings()` or something like that (I need to check it since I don't expect uniform API in different browsers).

PR #32, opened 2018-06-05
react native handler correction

ajaybhatt17:


lib/handlers/ReactNative.js
```diff
@@ -56,6 +56,15 @@ class Handler extends EnhancedEventEmitter
...
 		});
+
+		this._pc.onnegotiationneeded = () =>
+		{
+			if (isOffer)
+			{
+				this.createOffer();
+			}
+		};
+
```
ibc: What is this `createOffer()` for?

lib/handlers/ReactNative.js
```diff
@@ -66,13 +75,26 @@ class Handler extends EnhancedEventEmitter
...
+		this._pc.createOffer().then((desc) =>
+		{
+			return this._pc.setLocalDescription(desc);
+		})
+			.catch((err) =>
+			{
+				logger.error(err);
+			});
+	}
+
```
ibc: I don't understand the purpose of this.

ibc:
I don't understand the purpose of this PR. To be clear: I have `mediasoup-client` working fine in `react-native` by using a modified version of `react-native-webrtc` (just the commit I recently shown in the mediasoup mailing list).

Not sure the rationale behind the `onnegotiationeeded`.

Also, the PR does not follow the `mediasoup-client` code syntax. But, before that, I need to understand why this PR is needed at all and what it is supposed to provide or fix.

ajaybhatt17:
The changes is for handling remote peers video which is not showing. 

Sorry if i missed, which commit you are refering ... let me try that one also.

And above two snippet code (onnegotiationeeded and createOffer) are refer from https://github.com/oney/RCTWebRTCDemo/blob/master/main.js

ibc:
> The changes is for handling remote peers video which is not showing.

Sorry, that's not needed at all and, in fact, breaks mediasoup-client usage.


> Sorry if i missed, which commit you are refering ... let me try that one also.

This: https://groups.google.com/d/msg/mediasoup/107-9cb98zE/NHCj6CyBCAAJ


> And above two snippet code (onnegotiationeeded and createOffer) are refer from https://github.com/oney/RCTWebRTCDemo/blob/master/main.js

Just because a `react-native-webrtc` demo project uses that it does not mean that every WebRTC usage with `react-native-webrtc` requires that. `mediasoup-client` does not work that way at all.

Sorry, I cannot accept this PR.

ajaybhatt17:
Yes you are right.. tried the mentioned code in react-native-webrtc, it is working. Sorry for your trouble.

Thanks

ibc:
Great :)

PR #33, opened 2018-06-06
Add iceTransportPolicy option to Room settings

james-maloney:
I currently need this option to change iceTransportPolicy from 'all' to 'relay'.

### DIFF
--- a/lib/handlers/Chrome55.js
+++ b/lib/handlers/Chrome55.js
@@ -20,7 +20,7 @@ class Handler extends EnhancedEventEmitter
 		this._pc = new RTCPeerConnection(
 			{
 				iceServers         : settings.turnServers || [],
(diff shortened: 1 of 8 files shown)


lib/handlers/Chrome55.js
```diff
@@ -20,7 +20,7 @@ class Handler extends EnhancedEventEmitter
 		this._pc = new RTCPeerConnection(
 			{
 				iceServers         : settings.turnServers || [],
-				iceTransportPolicy : 'all',
+				iceTransportPolicy : settings.iceTransportPolicy || 'all',
```
ibc: Let's remove the `|| 'all'` here and in all the handlers. It's not needed since `this._settings` in `Room.js` already has it as default value.

ibc:
Thanks! mediasoup-client 2.1.0 published.

PR #37, opened 2018-06-25
Update ICE gathering in MS Edge.

havfo:
Update ICE gathering according to: https://msdn.microsoft.com/en-us/library/mt433085(v=vs.85).aspx


havfo:
Seems like I might have misunderstood MS documentation. This function only retrieves candidates allready discovered. Does not start gathering?

ibc:
This is not correct. Yes, Edge lacks the gather() method but that doesn't mean that getLocalCandidates() should be called instead.

PR #41, opened 2018-08-10
add sdpSemantics: 'plan-b'

fippo:
to make the plan-b choice independent of Chromes default behaviour.


ibc:
Yeah

marchoh:
Hi @ibc,
Is it possible to add this to the 2.* branch as well?
Since Chrome 72 its needs to be explicitly set.
When I set it manually, it fixes some issues I have (not getting a remote picture from Firefox within Chrome)

See here:
https://webrtc.org/web-apis/chrome/unified-plan/

ibc:
If there is a bug please report it. Forcing "plan-b" in Chrome 72 is not the way to go. mediasoup-client uses Unified-Plan in Chrome >= 70.

PR #45, opened 2018-10-22
use debug dist

artushin:
debug 4.1.0's default entrypoint isn't es5

### DIFF
--- a/lib/Logger.js
+++ b/lib/Logger.js
@@ -1,4 +1,4 @@
-import debug from 'debug';
+import debug from 'debug/dist/debug';
 
 const APP_NAME = 'mediasoup-client';
 


ibc:
Hi, I know that. But I had a heavy discussion with the `debug` author and finally accepted that the world is moving to ES6 even for minified bundles.

The way to go is using a proper ES6 minifier in your app, such as [terser](https://www.npmjs.com/package/terser). Note that `gulp-uglify-es` also uses `terser` internally.

Usage example: https://github.com/versatica/mediasoup-demo/blob/master/app/gulpfile.js#L28

skystebnicki:
It’s a huge bummer that this got rejected. It means anyone who still has to support IE11 users is SOL and cannot use your library. Why not support it?

ibc:
Two points:

1. IE11 does not support WebRTC nor ORTC, so anyway `mediasoup-client` does not work in IE11.
1. The fact that `mediasoup-client` uses ES6 modules does not mean that you need to run ES6 code in your JS app. It's up to you to transpile your whole code app (which includes `mediasoup-client` as a dependency) to ES5 and how you bundle it into a single JS file. You can use browserify / webpack / babel and use their options to convert ES6 to ES5, including dependencies.

PR #46, opened 2018-10-24
a jest testing poc

ivmos:
Hi friends

I've tried to do some basic tests for mediasoup-client, it can be challenging and fun because of the way it's architected.

Probably this PR is not good enough for merging, I might even break something as I needed to mess with deps, but I wanted to show what I tried and this serves as PR++ for Hacktoberfest even if not accepted. Perhaps it can help as reference if adding unit tests is scoped in the future.

Thank you and happy Hacktoberfest :)

captura de pantalla 2018-10-24 a las 22 21 46

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -3,5 +3,5 @@
 	[
 		[ "transform-builtin-classes", { "globals": [ "Error" ] } ]
 	],
-	"presets": [ "@babel/env" ]
+	"presets": [ "env" ]
 }
--- a/__tests__/Room.test.js
+++ b/__tests__/Room.test.js
@@ -1,0 +1,65 @@
+const mediasoupClient = require('../lib/index')
+const { TimeoutError } = require('../lib/errors')
+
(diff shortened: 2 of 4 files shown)


ibc:
Hi, this is definitely a must. We need test units in mediasoup-client.

Said that, it's important to test the mediasoup-client surface API, but it's even more important to test that the underlying stuff (per-browser specific handlers) do work as expected.

Let's discuss about this next week with tons of beers.

ibc:
Hi, we are planning and defining [mediasoup v3](https://github.com/versatica/mediasoup/issues/227). In both, client and server sides, v3 will expose a much more granular API that will make it possible to add more test units as those you propose here.

Will come back to this once v3 is done.

ibc:
Hey, I've added `jest` in mediasoup-client v3 branch: e43dd31624ac031968f5afb9fd9a9a147bc0d0b

So thanks very useful :)

ivmos:
Great thanks!

Issue #5, opened 2017-11-06
Support react-native-webrtc


ibc:
* https://github.com/oney/react-native-webrtc

ibc:
Working on it: db3528dcbdb477764faba241548570c34cb1db1b

ibc:
First problem when in iOS: https://github.com/oney/react-native-webrtc/issues/401

ibc:
So the first problem may have been bypassed with this commit: d77f4a48673114f815f538599b75ab752752b7bc

thehappycoder:
@ibc Do you have video publishing working now that you have https://github.com/ibc/libsdptransform or there is still more work to be done to get first green path?

thehappycoder:
Can you share the source code for the demo you are building? After adding https://github.com/versatica/mediasoup-client.git#react-native, protoo-client, and events packages, I connected to the websocket and did `room.join`, which crashed with an error:
ReferenceError: Can't find variable: RTCPeerConnection because such class doesn't exist in react-native land. I don't see you importing it from `react-native-webrtc`

ibc:
libsdptransform is needed bt the future mediasoup-client-native (built in C++). This task is about using react native plus react-native-webrtc, so mediasoup-client (JS) itself should work. Of course, it needs react-native-webrtc.
We are working on it.

ibc:
However, let's keep please this issue clean with just comments about its progress.

ibc:
Another big limitation in `react-native-webrtc`: [Cannot use a new create MediaStream to display a local or remote video track](https://github.com/oney/react-native-webrtc/issues/405)

Basically `react-native-webrtc` assumes that the only `stream` objects that the app is gonna use are those `streams` generated by `react-native-webrtc` (via `getUserMedia` or `setRemoteDescription`).

This is a heavy limitation for `mediasoup-client` which is not designed to depend on the original `MediaStream` at all. So bad news.

ibc:
NOTE: **mediasoup-client 2.0.7** works with [react-native-webrtc](https://github.com/oney/react-native-webrtc/).

nxz91:
Hi @ibc, when you say it works you mean it works without modifications except video tracks? Audio connections should be fine?

nxz91:
A few fun facts:

```  "dependencies": {
...,
    "mediasoup-client": "^2.0.16",
    "react": "16.2.0",
    "react-native": "0.52.0",
    "react-native-webrtc": "^1.58.3",
  }```

[...]

PR #54, opened 2018-11-06
Make mediasoup-client compatible with react-native-webrtc

Sharcoux:
extends MediaStream to make mediasoup-client fully and immediately compatible with react-native-webrtc

### DIFF
--- a/lib/handlers/ReactNative.js
+++ b/lib/handlers/ReactNative.js
@@ -9,6 +9,7 @@ import RemotePlanBSdp from './sdp/RemotePlanBSdp';
 
 // We try to load react-native-webrtc library that should be available if we are using this handler.
 let webrtc = {};
+const currentPeerConnections = new Map();
 
 try
 {
@@ -18,6 +19,27 @@ try
 	
 	// We will need to override the addTrack method of the MediaStream object.
 	MediaStream.prototype.oldAddTrack = MediaStream.prototype.addTrack;
+
+	MediaStream.prototype.addTrack = function(track)
+	{
+		// If the track doesn't come from an existing stream, we use the current stream tag as
+		// streamReactTag, or we create one if it doesn't exist yet
+		if (!track.streamReactTag) 
+		{
+			track.streamReactTag = this.reactTag || Date.now();
+		}
+		// If the current stream doesn't have a react tag, we use the stream tag of the track
+		// we are adding. This will make us able to extract a track from a stream.
+		if (!this.reactTag) 
+		{
+			this.reactTag = track.streamReactTag;
+			this.id = this.reactTag;
+		}
+		// We call the original function.
+		this.oldAddTrack(track);
+		// We make sure that the current stream is accessible by every RTCPeerConnection.
+		currentPeerConnections.forEach((pc) => pc.addStream(this));
+	};
 }
 catch (err) {}
 
@@ -38,31 +60,10 @@ class Handler extends EnhancedEventEmitter
 				bundlePolicy       : 'max-bundle',
 				rtcpMuxPolicy      : 'require'
 			});
-
-		const self = this;
+		
 		// Each time we add a track to a MediaStream, we need to add it to the list of streams
 		// accessible by the RTCPeerConnection provided by react-native-webrtc
-
-		MediaStream.prototype.addTrack = function(track) 
-		{
-			// If the track doesn't come from an existing stream, we use the current stream tag as
-			// streamReactTag, or we create one if it doesn't exist yet
-			if (!track.streamReactTag) 
-			{
-				track.streamReactTag = this.reactTag || Date.now();
-			}
-			// If the current stream doesn't have a react tag, we use the stream tag of the track
-			// we are adding. This will make us able to extract a track from a stream.
-			if (!this.reactTag) 
-			{
-				this.reactTag = track.streamReactTag;
-				this.id = this.reactTag;
-			}
-			// We call the original function.
-			this.oldAddTrack(track);
-			// We make sure that the current stream is accessible by the RTCPeerConnection.
-			self._pc.addStream(this);
-		};
+		currentPeerConnections.set(this, this._pc);
 
 		// Generic sending RTP parameters for audio and video.
 		// @type {Object}
@@ -100,6 +101,7 @@ class Handler extends EnhancedEventEmitter
 	close()
 	{
 		logger.debug('close()');
+		currentPeerConnections.delete(this);
 
 		// Close RTCPeerConnection.
 		try { this._pc.close(); }
--- a/lib/handlers/ReactNative.js
+++ b/lib/handlers/ReactNative.js
@@ -9,36 +9,37 @@ import RemotePlanBSdp from './sdp/RemotePlanBSdp';
 
 // We try to load react-native-webrtc library that should be available if we are using this handler.
 let webrtc = {};
-const currentPeerConnections = new Map();
 
 try
 {
 	webrtc = require('react-native-webrtc');
 	// We make react-native-webrtc API available from the global object.
 	Object.assign(global, webrtc);
-	
-	// We will need to override the addTrack method of the MediaStream object.
-	MediaStream.prototype.oldAddTrack = MediaStream.prototype.addTrack;
 
-	MediaStream.prototype.addTrack = function(track)
+	// We extend MediaStream class.
+	global.MediaStream=class MediaStream extends webrtc.MediaStream {};
+
+	// Streams containing video need to track the stream registered in the peerConnection and denoted by 
+	// track.streamReactTag.
+	global.MediaStream.addTrack = function(track) 
 	{
-		// If the track doesn't come from an existing stream, we use the current stream tag as
-		// streamReactTag, or we create one if it doesn't exist yet
-		if (!track.streamReactTag) 
-		{
-			track.streamReactTag = this.reactTag || Date.now();
-		}
-		// If the current stream doesn't have a react tag, we use the stream tag of the track
-		// we are adding. This will make us able to extract a track from a stream.
-		if (!this.reactTag) 
+		if (track.kind === 'video') 
 		{
 			this.reactTag = track.streamReactTag;
 			this.id = this.reactTag;
 		}
-		// We call the original function.
-		this.oldAddTrack(track);
-		// We make sure that the current stream is accessible by every RTCPeerConnection.
-		currentPeerConnections.forEach((pc) => pc.addStream(this));
+		super.addTrack(track);
+	};
+
+	// Streams created with getUserMedia should contain a streamReactTag.
+	global.getUserMedia=function(constraints, success, error) 
+	{
+		return webrtc.getUserMedia(constraints, success, error).then((stream) => 
+		{
+			stream.getTracks().forEach((track) => (track.streamReactTag=stream.reactTag));
+			
+			return stream;
+		});
 	};
 }
 catch (err) {}
@@ -61,10 +62,6 @@ class Handler extends EnhancedEventEmitter
 				rtcpMuxPolicy      : 'require'
 			});
 		
-		// Each time we add a track to a MediaStream, we need to add it to the list of streams
-		// accessible by the RTCPeerConnection provided by react-native-webrtc
-		currentPeerConnections.set(this, this._pc);
-
 		// Generic sending RTP parameters for audio and video.
 		// @type {Object}
 		this._rtpParametersByKind = rtpParametersByKind;
@@ -101,7 +98,6 @@ class Handler extends EnhancedEventEmitter
 	close()
 	{
 		logger.debug('close()');
-		currentPeerConnections.delete(this);
 
 		// Close RTCPeerConnection.
 		try { this._pc.close(); }
--- a/lib/handlers/ReactNative.js
+++ b/lib/handlers/ReactNative.js
@@ -7,18 +7,19 @@ import * as sdpCommonUtils from './sdp/commonUtils';
 import * as sdpPlanBUtils from './sdp/planBUtils';
 import RemotePlanBSdp from './sdp/RemotePlanBSdp';
 
-//We try to load react-native-webrtc library that should be available if we are using this handler
+// We try to load react-native-webrtc library that should be available if we are using this handler.
 let webrtc = {};
-try {
-  webrtc = require('react-native-webrtc');
-} catch (err) {}
 
-//We make react-native-webrtc API available from the global object
-Object.assign(global, webrtc);
-
-//We will need to override the addTrack method of the MediaStream object
-MediaStream.prototype.oldAddTrack = MediaStream.prototype.addTrack;
-
+try
+{
+	webrtc = require('react-native-webrtc');
+	// We make react-native-webrtc API available from the global object.
+	Object.assign(global, webrtc);
+	
+	// We will need to override the addTrack method of the MediaStream object.
+	MediaStream.prototype.oldAddTrack = MediaStream.prototype.addTrack;
+}
+catch (err) {}
 
 const logger = new Logger('ReactNative');
 
@@ -39,24 +40,27 @@ class Handler extends EnhancedEventEmitter
 			});
 
 		const self = this;
-		//Each time we add a track to a MediaStream, we need to add it to the list
-		//of streams accessible by the RTCPeerConnection provided by react-native-webrtc
-		MediaStream.prototype.addTrack = function(track) {
-			//If the track doesn't come from an existing stream, we use the current
-			//stream tag as streamReactTag, or we create one if it doesn't exist yet
-			if (!track.streamReactTag) {
+		// Each time we add a track to a MediaStream, we need to add it to the list of streams
+		// accessible by the RTCPeerConnection provided by react-native-webrtc
+
+		MediaStream.prototype.addTrack = function(track) 
+		{
+			// If the track doesn't come from an existing stream, we use the current stream tag as
+			// streamReactTag, or we create one if it doesn't exist yet
+			if (!track.streamReactTag) 
+			{
 				track.streamReactTag = this.reactTag || Date.now();
 			}
-			//If the current stream doesn't have a react tag, we use the stream tag
-			//of the track we are adding. This will make us able to extract a track
-			//from a stream.
-			if (!this.reactTag) {
+			// If the current stream doesn't have a react tag, we use the stream tag of the track
+			// we are adding. This will make us able to extract a track from a stream.
+			if (!this.reactTag) 
+			{
 				this.reactTag = track.streamReactTag;
 				this.id = this.reactTag;
 			}
-			//We call the original function
+			// We call the original function.
 			this.oldAddTrack(track);
-			//We make sure that the current stream is accessible by the RTCPeerConnection
+			// We make sure that the current stream is accessible by the RTCPeerConnection.
 			self._pc.addStream(this);
 		};
 
--- a/lib/handlers/ReactNative.js
+++ b/lib/handlers/ReactNative.js
@@ -7,6 +7,19 @@ import * as sdpCommonUtils from './sdp/commonUtils';
 import * as sdpPlanBUtils from './sdp/planBUtils';
 import RemotePlanBSdp from './sdp/RemotePlanBSdp';
 
+//We try to load react-native-webrtc library that should be available if we are using this handler
+let webrtc = {};
+try {
+  webrtc = require('react-native-webrtc');
+} catch (err) {}
+
+//We make react-native-webrtc API available from the global object
+Object.assign(global, webrtc);
+
+//We will need to override the addTrack method of the MediaStream object
+MediaStream.prototype.oldAddTrack = MediaStream.prototype.addTrack;
+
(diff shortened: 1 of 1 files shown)


ibc:
Sorry, I cannot accept this PR:

* It does not follow the coding style of mediasoup-client (`gulp lint` does not pass).
* It pollutes `window` object, so it's intrusive.
* This issue must be fixed in `react-native-webrtc` project, which obviously implements a super outdated and limited WebRTC API (still stream based), and makes wrong assumptions such as assuming that the only MediaStreams that the app must use are those generated by the `react-native-webrtc`.

I plan to work on this in the future, but I've no time/resources right now.

Sharcoux:
I agree, but at least there is a trace of a working solution. I think that most of users would rather that than a potential rewriting of react-native-webrtc in an improbable future. Moreover, the code in the ReactNative file is executed only for React Native users, so I wonder if it is not better to propose them with a solution that just work...
Anyway, it is your project. Do as you wish.
About lint, my bad. Why not adding a precommit hook though? Lint-staged + husky does the job pretty well.

Sharcoux:
> Syntax (this included comments, which must have space after //, blank line before, and must end with dot.

Ok

> It assumes that there is require() in runtime. This won't work if someone uses the dist/mediasoup-client.js bundle.

This is the point. The ReactNative handler will be used only with React-Native projects thanks to [this](https://github.com/versatica/mediasoup-client/blob/master/lib/Device.js#L110) (at least, that's how I understand this code). Such a project would need to include react-native-webrtc of course. But we need to use a dynamic import to keep `mediasoup` compatible with web projects that don't need `react-native-webrtc`.

> It adds MediaStream.prototype.oldAddTrack always (this and the Object.assign should be within an if block just if react-native-webrtc has been detected.

Yeah, why not, but anyway, the code won't work if the dependency has not been found so... But I must agree with you on this one.

> It completely ignores the already working mechanism that mediasoup-client includes to detect react-native environment: this.

I don't understand what you mean. I based this on the assumption that the ReactNative handler is used only in React-Native projects.

> It uses const self = this instead of arrow functions.

I don't see how you could do better as you need both `this` as the `MediaStream` object and `this` as the `Handler`

> within the Handler class constructor it overrides the addTrack method to MediaStream class. This executed everytime a Handler instance is created. This should be executed just once (at the beginning if react-native-webrtc has been detected).

Unfortunately, not. You need a reference to the `_pc` object which is instantiated inside the Handler

Sharcoux:
Just to be clear, I'm not saying that what I propose here is great. It's hacky as hell. It is just that it is the only way that I found that will work with react-native-webrtc as it is right now.

By the way, I'm not really worried about polluting the global object with the objects from react-native-webrtc. It is more or less the point of that lib.

ibc:
Hi, I understand your concerns and you are right. I just want to avoid a PR does just makes mediasoup-client work with react-native-webrtc in a specific super simplified use case.

Specifically, `MediaStream.prototype.addTrack` is created within the `Handler` constructor because, as you said, it needs access to `this._pc`.  Well, I think this assumes too much. For example, my app uses different sending transports and not just one. Adding a track into a stream (regardless it's done internally by react-native-webrtc) should not assume that **any** mediaspoup sending transport should send such a track.

I don't remember the details right now, but I have an (unmaintained and outdated) fork of react-native-webrtc that works (or worked) with mediasoup-client, and the only change was this:

screenshot 2018-11-07 at 02 12 41

I'm ok with adding a "temporal" hack to mediasoup-client in the `ReactNative.js` handler to make it work, but I cannot assume your above assumption. Can we work in a different hack that, somehow, just does what my screenshot does?

Sharcoux:
You're right, but the issue with your change is that your hack is not enough. Basically, if you write `new MediaStream` anywhere in your code, it won't work. And I don't see how I can embed the video track and the audio track received by each consumer inside one stream without using `new MediaStream`.

I proposed a hack that would make any stream generated with `new MediaStream` available for every ReactNative handler.

ibc:
> You're right, but the issue with your change is that your hack is not enough.

Such a change plus the `lib/handlers/ReactNative.js` handler does work with the current mediasoup-demo.

> Basically, if you write new MediaStream anywhere in your code, it won't work.

It works because the video track you will insert into that new MediaStream already contains a `streamReactTag` (that what the commit above in react-native-webrtc does). Later, when creating a mediasoup-client Producer it internally handles a `new MediaStream(track.streamReactTag)`:

https://github.com/versatica/mediasoup-client/blob/master/lib/handlers/ReactNative.js#L109

so the stream will have a `reactTag` with that value (and that's what react-native-webrtc internally needs to identify the track to send):

https://github.com/oney/react-native-webrtc/blob/master/MediaStream.js#L39

And that stream is given to the `pc` via `addStream()`.


> And I don't see how I can embed the video track and the audio track received by each consumer inside one stream without using new MediaStream.

The mediasoup-client `ReactNative` handler already manages that by adding the corresponding `streamReactTag` to the returned `track`. If you apply the patch above and insert the Consumer returned `track` into a new MediaStream, you can use such a stream into a react-native-webrtc `RTCView`.

> I proposed a hack that would make any stream generated with new MediaStream available for every ReactNative handler.

What your hack does is calling `pc.addStream(stream)` into all the existing peerconnections when `stream.addTrack(track)` is called. Adding a track into a stream should not involve also adding such a stream into all existing peerconnections.

Sharcoux:
If you have streamA containing audio track, and streamV containing video track, you need to create streamAV containing audio **and** video. The issue is that streamAV will not be accessible by react-native-webrtc,

For what I remember, we need this:

		if (!track.streamReactTag) 
		{
			track.streamReactTag = this.reactTag || Date.now();
		}

To make us able to display our localStream as feedback. Otherwise, the track that you get through getUserMedia will not have a streamReactTag set and won't be rendered by the RTCView

We need this:

      if (!this.reactTag) {
        this.reactTag = track.streamReactTag;
        this.id = this.reactTag;
      }

In order to be able to create a composite stream from multiple tracks. Otherwise, the stream you created will not be registered and won't be rendered by the RTCView. You could create a new reactTag for your new stream, but in this case, you won't be able to use the stream your track comes from. That's why the best solution I could find was to use the streamReactTag of the track.

We need this:

		// We make sure that the current stream is accessible by every RTCPeerConnection.
		currentPeerConnections.forEach((pc) => pc.addStream(this));

To make sure that you will be able to display your MediaStream in your RTCView. It is just a map anyway, so it costs almost nothing to just add every stream that contains tracks to every peerConnection. I don't see a better way.

ibc:
> If you have streamA containing audio track, and streamV containing video track, you need to create streamAV containing audio **and** video. The issue is that streamAV will not be accessible by react-native-webrtc
> 
> For what I remember, we need this:
> 
> ```
> 	if (!track.streamReactTag) 
> 	{
> 		track.streamReactTag = this.reactTag || Date.now();
> 	}
> ```
> To make us able to display our localStream as feedback. Otherwise, the track that you get through getUserMedia will not have a streamReactTag set and won't be rendered by the RTCView

Tracks are always generated by react-native-webrtc, so if my patch above is applied into react-native-webrtc, those tracks already have `streamReactTag`.

 
> We need this:
> 
> ```
> 	// We make sure that the current stream is accessible by every RTCPeerConnection.
> 	currentPeerConnections.forEach((pc) => pc.addStream(this));
> ```
> To make sure that you will be able to display your MediaStream in your RTCView. It is just a map anyway, so it costs almost nothing to just add every stream that contains tracks to every peerConnection. I don't see a better way.

Not sure if I explained myself properly above, but want to make clear that, with the patch above (attached in a image) and the current `ReactNative.js` handler, I don't need such a hack.

Anyway, again: we cannot assume that any track added into a stream means that it should be added into one or many peerconnections.

ibc:
> > Tracks are always generated by react-native-webrtc, so if my patch above is applied into react-native-webrtc, those tracks already have streamReactTag.
> 
> Aren't you doing the same kind of assumption as the one of react-native-webrtc that lead us here? ;)
> What about the tracks that have been generate with `getUserMedia` on the mobile for instance? These tracks won't have a streamReactTag and thus, the RTCView can't display them (think about a feedback of your local stream).

If you use React-Native without `react-native-webrtc` you don't have `getUserMedia`. If you have `react-native-webrtc` the you use the `getUserMedia` provided by it, which internally generates JS track objects.
 
> > Not sure if I explained myself properly above, but want to make clear that, with the patch above (attached in a image) and the current ReactNative.js handler, I don't need such a hack.
> 
> Was your patch merged into react-native-webrtc? If not, don't you think it would be better to patch what you have control upon?

Honestly: because I don't want to be responsible of a little and limited hack I just did as a quick proof of concept for a demo app. Feel free to propose that PR in the project :)
 
> > Anyway, again: we cannot assume that any track added into a stream means that it should be added into one or many peerconnections.
> 
> Why not? It costs almost nothing, and at least RTCView will be able to display them. What would be the point of having streams that you cannot play ?

May be I miss something here. Do you mean that, when using `react-native-webrtc` you cannot use `getUserMedia()` and play the local video track into a `RTCView` **without** also running a `peerconnection` and adding such a stream into it? /cc @saghul.

Sharcoux:
> If you use React-Native without react-native-webrtc you don't have getUserMedia. If you have react-native-webrtc the you use the getUserMedia provided by it, which internally generates JS track objects.

Only with your patch.

> Honestly: because I don't want to be responsible of a little and limited hack I just did as a quick proof of concept for a demo app. Feel free to propose that PR in the project :)

You did NOT ??? Well... I should as well start with that. I thought that react-native-webrtc refused to insert your patch. But yet, it is not enough as you also need to patch streams created with `new MediaStream``

> May be I miss something here. Do you mean that, when using react-native-webrtc you cannot use getUserMedia() and play the local video track into a RTCView without also running a peerconnection and adding such a stream into it? /cc @saghul.

Well... You got me, here... ^^', actually, about this part, the concern is more about being able to use the same component for every stream, being local or remote. But I agree that it is not required.

ibc:
> > If you use React-Native without react-native-webrtc you don't have getUserMedia. If you have react-native-webrtc the you use the getUserMedia provided by it, which internally generates JS track objects.
> 
> Only with your patch.

Yes, of course. All I said was assuming my little patch is applied to `react-native-webrtc`. My point was that you said:

> What about the tracks that have been generate with getUserMedia on the mobile for instance? 

But that does not make sense in React-Native (there is no `getUserMedia` other than the one provided by `react-natite-webrtc`).


> > Honestly: because I don't want to be responsible of a little and limited hack I just did as a quick proof of concept for a demo app. Feel free to propose that PR in the project :)
> 
> You did NOT ??? Well... I should as well start with that. I thought that react-native-webrtc refused to insert your patch. But yet, it is not enough as you also need to patch streams created with `new MediaStream`

Not exactly. You need to create those streams by passing as argument `streamReactTag` of the corresponding video `track` that later you are gonna insert into that new stream. This is what mediasoup-client does:

https://github.com/versatica/mediasoup-client/blob/master/lib/handlers/ReactNative.js#L109

And yes, such a hack should be documented if my little patch is accepted.

ibc:
> Not exactly. You need to create those streams by passing as argument the `streamReactTag` of the corresponding video `track` that later you are gonna insert into that new stream. This is what mediasoup-client does:
> 
> https://github.com/versatica/mediasoup-client/blob/master/lib/handlers/ReactNative.js#L109

A better approach would be including in the PR an addition into `MediaStream.prototype.addTrack`:

If `track.kind === "video"` then update `this.reactTag` with the value of `track.streamReactTag)`.

Super ugly, yes, as it assumes a single video track per stream (more than one would not work anyway), but at least it does not require extra doc.

Sharcoux:
Ok ! Ok ! I think that I found the perfect way to patch it and make everyone happy. Let me show you in my next commit.

Sharcoux:
Ok. I didn't have time to test yet, but I think that you might like better this approach.

I still have a few doubts, though:
- Why do you need the reactTag only for video tracks ? Don't we need audio too?
- I don't understand what is the point of your patch. You set the streamReactTag for the tracks generated by getUserMedia, but such tracks are never added to the peerConnection anyway, so they won't be rendered by a RTCView. I'm probably missing sth here, but what is the point?

ibc:
> * Why do you need the reactTag only for video tracks ? Don't we need audio too?

That question corresponds to the `react-native-webrtc` project. I don't like how it's designed at all and I don't want to depend on a wrong API.

> * I don't understand what is the point of your patch. You set the streamReactTag for the tracks generated by getUserMedia, but such tracks are never added to the peerConnection anyway, so they won't be rendered by a RTCView. I'm probably missing sth here, but what is the point?

I already explained that, in order to play a video in a `RTCView` you **DO NOT** need to add its stream into any peerconnection.

----

Now, coming back to this PR, if you think about it it, right now your PR just monkey patches the `react-native-webrtc` code. It's not related to mediasoup-client anymore. So the question is:

* Why don't you propose a PR into `react-native-webrtc` instead?
* I don't want to maintain code in `mediasoup-client` that depends on a specific version (a limited and buggy version) of `react-native-webrtc` and I don't want to monkey patch external libraries.
* I don't want to document anywhere or explain users that "hey, if you want to use mediasoup-client with react-native-webrtc then you need to use `global.getUserMedia` instead of `webrtc.getUserMedia`".

ibc:
It's clear IMHO that, finally, changes being proposes in this PR are just monkey patches to `react-native-webrtc`. If we want to be able to manage locally created streams (using `new MediaStream()`) those changes must be apply in `react-native-webrtc` project rather than making a separate project monkey patch it.

Sharcoux:
> I don't like how it's designed at all and I don't want to depend on a wrong API.

Hum... Maybe, but right now there is no other way to use mediasoup-client on React-Native. So...

> I already explained that, in order to play a video in a RTCView you DO NOT need to add its stream into any peerconnection.

Actually you're partially right. I forgot something. streams generated by `getUserMedia` are actually added to the localStreams and can be displayed by an RTCView.

```
getVideoTrackForStreamURL
getStreamForReactTag
localStreams.get(streamReactTag)
webRTCModule.localStreams.put(streamId, mediaStream);
                PeerConnectionObserver pco
                    = mPeerConnectionObservers.valueAt(i);
                stream = pco.remoteStreams.get(streamReactTag);
```

Maybe making a PR on webrtc-react-native that would just add any stream created with new MediaStream to the local streams would be enough, but I don't think that I can do that if I add remote tracks into a local stream.

ibc:
Ok, let's clarify how react-native-webrtc works regarding video tracks:

* It generates streams with a `reactTag` identificator that is mapped internally.
* Such a `reactTag` is the only identification for the lib internals to know which video track to render.
* react-native-webrtc **wrongly** assumes that the app will just use streams generated by react-native-webrtc.

So, the hack that mediasoup-client does for remote video tracks is:

* Get the new remote stream generated by react-native-webrtc.
* Copy the `stream.reactTag` into `track.streamReactTag`:

```js
// Hack: Add a streamReactTag property with the reactTag of the MediaStream
// generated by react-native-webrtc (this is needed because react-native-webrtc
// assumes that we're gonna use the streams generated by it).
track.streamReactTag = stream.reactTag;
```

* And resolve the promise with such a track.

Later when the app wants to render such a video track:

* It must create a new stream whose `reactTag` matches the `track.streamReactTag`:

```js
const stream = new MediaStream(track.streamReactTag);

// now render such a stream into a RTCView.
```

It will work because when later doing `view.src = URL.createObjectUrl(stream)`, react-native-webrtc will read the `reactTag` of the stream (which we previously set to match `track.streamReactTag` and look internally for the corresponding video track.

This is, it will work because both the stream generated by react-native-webrtc and our local stream have the same `reactTag`. That's the **only** that is needed.

That's all. mediasoup-client already add such a `track.streamReactTag` internally for remote tracks (check `addConsumer()` method in `lib/handlers/ReactNative.js`) so the only missing part is also adding it to tracks generated by `getUserMedia`, which are exactly those 2-3 lines:

https://user-images.githubusercontent.com/16191/48104205-b7852a80-e232-11e8-9be8-4f9987d02fee.png

And then the app must always do:

```js
const stream = new MediaStream(track.streamReactTag);
```

Sharcoux:
Ok, forget everything I said. You were 100% right about everything.

[https://github.com/react-native-webrtc/react-native-webrtc/pull/628](https://github.com/react-native-webrtc/react-native-webrtc/pull/628)

ibc:
Great, hope it's accepted.

Sharcoux:
Note that the problem I had was that to display your stream, you have to do something like:

`    const url = stream.getTracks().length ? stream.getTracks()[0].streamReactTag : stream.toURL();
`

Or at least, set your stream reactTag to the streamReactTag of the track you're adding to it...

Sharcoux:
In the end, I wonder if it is not required to still change MediaStream for it to read the streamReactTag of the tracks we add to it...

ibc:
> In the end, I wonder if it is not required to still change MediaStream for it to read the streamReactTag of the tracks we add to it...

Honestly I do not remember it very well, but as far as I remember the `MediaStream` JS class in react-native-webrtc is just a fake container that does nothing and holds nothing.

Sharcoux:
No, it can hold tracks and we don't ask much more from it... Except to be able to provide that information to the RTCView which is what it totally fails at right now.

Sharcoux:
By the way, I don't understand how the sound tracks are being handled...

ibc:
> By the way, I don't understand how the sound tracks are being handled...

By default, the libwebrtc bindings for Android/iOS auto-magically handle audio tracks by directly playing them, which is ugly.

DimitrisTzimikas:
@Sharcoux have you been able to connect mediasoup-client with rn-webrtc? I've set a working example with -v 1.69.1, but I can't figure it out. If you have a demo it would be great. Thnks in advance.

Sharcoux:
I am currently using the same code for web and all devices. While resolving the PR into rn-webrtc, you can check the content of this current PR. It should be enough to make it work.

If it still doesn't, just tell me more about the errors you are facing.

jasonkang14:
@Sharcoux @ibc, I have checked [react-native-webrtc issue(#628)](https://github.com/react-native-webrtc/react-native-webrtc/pull/628) which is now closed after a PR from @Sharcoux got accepted. But just to make sure, is `mediasoup` compatible with `react-native-webrtc` now? If so, are both v2 and v3 compatible with `react-native-webrtc`? or just either one of the versions? 

@DimitrisTzimikas since you have set a working example, please let me know which version of `mediasoup` you used to make it work.

Sharcoux:
@jasonkang14 I made it work using Mediasoup V2 and the current PR

jasonkang14:
@Sharcoux Did you make any changes to `mediasoup` as well? and I am assuming `mediasoup v3` is not compatible with `react-native-webrtc` yet.

ibc:
> Did you make any changes to mediasoup as well? and I am assuming mediasoup v3 is not compatible with react-native-webrtc yet.

react-native-webrtc is as compatible with mediasoup v3 as it's with mediasoup v2. Whether it works or not depends on the hack being applied in react-native-webrtc sources.

With the new and amazing refactor @saghul is doing in react-native-webrtc, we'll no longer need such a hack. However we'll need to update the `ReactNative.js` handler in mediasoup-client when that happens.

Other than that, I don't know the current status and I'm not using react-native-webrtc lately so I cannot help much more for now.

saghul:
react-native-webrtc 1.75.0 is out, which includes support for the MediaStream constructor so the hack can be removed, as long as you use that version.

jasonkang14:
Appreciate your prompt responses

ibc:
Oh amazing. Which API does it expose? Stream based yet?

Anyway, does it mean that a MediaStream can be externally created, local tracks can added, and the stream can be given to pc.addStream()? And the same for remote tracks (create a new MediaStream, add remote tracks, and provide the stream to a video elem)?

saghul:
> Which API does it expose? Stream based yet?

Yes, still streams. Slowly but surely we'll arrive at transceivers :-)



> Anyway, does it mean that a MediaStream can be externally created, local tracks can added, and the stream can be given to pc.addStream()? And the same for remote tracks (create a new MediaStream, add remote tracks, and provide the stream to a video elem)?

Correct.

ibc:
Super great. So, in theory, I may remove the mediasoup-client ReactNative.js handler and use the Chrome56.js handler instead, right? It used plan-b and stream based API.

saghul:
I guess having a Rn specific one is good, because of the imports, at least.

ibc:
> I guess having a Rn specific one is good, because of the imports, at least.

Right now we do the same `imports` (`require` in fact) in both the ReactNative.js and Chrome55.js handlers:

https://github.com/versatica/mediasoup-client/blob/v3/lib/handlers/ReactNative.js?ts=2

ibc:
BTW we do this to detect RN:

```js
	if (typeof navigator === 'object' && navigator.product === 'ReactNative')
	{
		if (typeof RTCPeerConnection === 'undefined')
		{
			logger.warn('unsupported ReactNative without RTCPeerConnection');

			return null;
		}

		return ReactNative;
	}
```

Is there any way to know the react-native-webrtc version?

ibc:
BTW let's please move the discussion to https://github.com/versatica/mediasoup-client/issues/5#issuecomment-523476978 (as this is a closed PR).

PR #55, opened 2018-11-16
Custom device

jmillan:


lib/internals.js
```diff
@@ -0,0 +1,30 @@
...
+
+export {
+	Logger,
+	EnhancedEventEmitter,
+	utils,
+	ortc,
+	sdpCommonUtils,
+	sdpUnifiedPlanUtils,
+	sdpPlainRtpUtils,
+	sdpPlanBUtils,
```
ibc: Can we plz place sdpPlanBUtils before sdpPlainRtpUtils? (this was probably my fault)

jmillan: sure, done

PR #56, opened 2018-11-20
Don't use turns: urls with Edge.

michaelfig:
Microsoft Edge aborts the ICE gathering with `InvalidAccessError` if you supply a `turns:...` URL in the turnServers option.

Until Edge implements turns properly, downgrade all the `turns:...` URLs to just `turn:...`.

### DIFF
--- a/lib/handlers/Edge11.js
+++ b/lib/handlers/Edge11.js
@@ -325,9 +325,16 @@ export default class Edge11 extends EnhancedEventEmitter
 
 	_setIceGatherer(settings)
 	{
+		// NOTE: Edge halts the connection with InvalidAccessError when presented with a turns: server.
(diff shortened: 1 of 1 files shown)


ibc:
Thanks. However I don't want to add this kind of "magic" into mediasoup-client. Your PR assumes that if the given TURN servers are:

```
turns:example.org:443?transport=tcp
turn:example.org:6666?transport=tcp
turn:example.org:6666?transport=udp
```

then your PR transform them into this:

```
turn:example.org:443?transport=tcp  <--- This is incorrect
turn:example.org:6666?transport=tcp
turn:example.org:6666?transport=udp
```

Anyway, it's the app the responsible of not delivering TURN TLS servers to Edge.

PR #74, opened 2019-03-27
Handle Safari 12.1 sdp semantics

leader22:
Safari 12.1 may use `unified-plan` sdp and also `plan-b` sdp.
It depends on user settings.

In stable release, default sdp semantics is not a `unified-plan`.
In Technology Preview, `unified-plan` is default.

To handle it correctly, we need to check it in `detectBrowser`.

> https://webkit.org/blog/8672/on-the-road-to-webrtc-1-0-including-vp8/

(+ `eslint` does not have `RTCRtpTransciver` as its global variables.)

Feel free to close this if you already working on it. 👌

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -7,6 +7,10 @@ module.exports =
 		node: true,
 		'jest/globals': true
 	},
+	globals:
+	{
+		RTCRtpTransceiver: true
+	},
 	plugins:
 	[
 		'jest'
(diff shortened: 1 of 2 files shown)


ibc:
Thanks. I've done it in a very similar way in both mediasoup-client "v2" branch (NPM 2.4.10 released) and in "master" branch (incoming v3 version).

PR #85, opened 2019-08-30
ReactNative Handler modified after installing react-native-webrtc lib…

jasonkang14:
@ibc I am not sure where there is an error while building a project in React Native because it worked perfectly when I tried your example with Chrome. 

If you don't require `RTCPeerConnection`, `RTCSessionDescription`, and `MediaStream`, it throws an error saying that those three are not defined. 

Since you have worked hard to make `mediasoup` compatible with `react-native-webrtc`, I required `RTCPeerConnection`, `RTCSessionDescription`, and `MediaStream` from `react-native-webrtc` and got it to work. Please let me know if you think there is a better way to deal with the problem. 

RTCPeerConnection Error Message


jasonkang14:
I will change the name of the branch and make a new PR

ibc (commented):
No. mediasoup-client won't have react-native-webrtc as dependency.

PR #86, opened 2019-08-30
ReactNative: Requiring RTCPeerConnection, RTCSessionDescription, and MediaStream

jasonkang14:
@ibc I am not sure where there is an error while building a project in React Native because it worked perfectly when I tried your example with Chrome.

If you don't require RTCPeerConnection, RTCSessionDescription, and MediaStream, it throws an error saying that those three are not defined.

Since you have worked hard to make mediasoup compatible with react-native-webrtc, I required RTCPeerConnection, RTCSessionDescription, and MediaStream from react-native-webrtc and got it to work. Please let me know if you think there is a better way to deal with the problem.

RTCPeerConnection error

### DIFF
--- a/lib/handlers/ReactNative.js
+++ b/lib/handlers/ReactNative.js
@@ -7,6 +7,11 @@ const ortc = require('../ortc');
 const sdpCommonUtils = require('./sdp/commonUtils');
 const sdpPlanBUtils = require('./sdp/planBUtils');
 const RemoteSdp = require('./sdp/RemoteSdp');
+const {
+	RTCPeerConnection,
+	RTCSessionDescription,
+	MediaStream
+} = require('react-native-webrtc');
 
 const logger = new Logger('ReactNative');
 
--- a/package.json
+++ b/package.json
@@ -41,6 +41,7 @@
     "jest": "^24.9.0",
     "jest-tobetype": "^1.2.3",
     "node-mediastreamtrack": "0.1.0",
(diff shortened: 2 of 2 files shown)


ibc:
mediasoup-client won't include the whole react-native-webrtc as dependency, that would be terribly inappropriate.

@saghul can you help please? The problem seems to be that mediasoup-client depends on `RTCPeerConnection`, `RTCSessionDescription` and `MediaStream` in the global namespace.

* mediasoup-client **cannot** (or should not) call `require("react-native-webrtc")` internally.
* Does react-native-webrtc expose some API to pollute the global namespace?
* Or perhaps the [ReactNative.js](https://github.com/versatica/mediasoup-client/blob/v3/lib/handlers/ReactNative.js) handler could dynamically use `const { RTCPeerConnection } = require("react-native-webrtc")`, etc when needed? (so it would throw just in runtime if react-native-webrtc is not found)?
   * Is it a good practice to depend on runtime `require()` semantics within React-Native? or is it a super hack?

Other than that, any suggestion on what the best way to go is? A hack coming to my mind:

* Document that the application must require "react-native-webrtc" and pollute the global namespace before it creates a mediasoup-client `Device` instance.

ibc:
I'll close this PR. Let's continue the discussion in the already existing issue about react-native-webrtc:

https://github.com/versatica/mediasoup-client/issues/5#issuecomment-526581786

PR #87, opened 2019-09-02
MS Edge 18 support

milen-yordanov:
Works with MS Edge 18 after these small changes.
The missing codec.name  causes InvalidAccessError in Edge.

### DIFF
--- a/lib/handlers/Edge11.js
+++ b/lib/handlers/Edge11.js
@@ -182,9 +182,10 @@ class Edge11 extends EnhancedEventEmitter
 		await rtpSender.send(edgeRtpParameters);
 
 		// Store it.
+		const localId = track.id;
 		this._rtpSenders.set(track.id, rtpSender);
 
-		return rtpParameters;
+		return { localId, rtpParameters };
 	}
 
 	async stopSending({ localId })
--- a/lib/handlers/Edge11.js
+++ b/lib/handlers/Edge11.js
@@ -181,8 +181,9 @@ class Edge11 extends EnhancedEventEmitter
 
 		await rtpSender.send(edgeRtpParameters);
 
-		// Store it.
 		const localId = track.id;
+		
+		// Store it.
 		this._rtpSenders.set(track.id, rtpSender);
 
 		return { localId, rtpParameters };
--- a/lib/handlers/ortc/edgeUtils.js
+++ b/lib/handlers/ortc/edgeUtils.js
@@ -75,6 +75,11 @@ exports.mangleRtpParameters = function(rtpParameters)
 			delete codec.channels;
 		}
 
+		if(codec.mimeType && !codec.name)
+		{
+			codec.name = codec.mimeType.split('/')[1]
+		}
+
 		// Remove mimeType.
 		delete codec.mimeType;
 	}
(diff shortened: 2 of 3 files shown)


lib/handlers/ortc/edgeUtils.js
```diff
@@ -75,6 +75,11 @@ exports.mangleRtpParameters = function(rtpParameters)
 			delete codec.channels;
 		}
 
+		if(codec.mimeType && !codec.name)
+		{
+			codec.name = codec.mimeType.split('/')[1]
```
ibc: We don't use `codec.name` at all in the code but just  `codec.mimeType`. So this is not needed.

lib/handlers/Edge11.js
```diff
@@ -182,9 +182,10 @@ class Edge11 extends EnhancedEventEmitter
 		await rtpSender.send(edgeRtpParameters);
 
 		// Store it.
+		const localId = track.id;
 		this._rtpSenders.set(track.id, rtpSender);
 
-		return rtpParameters;
+		return { localId, rtpParameters };
```
ibc, line 189: Thanks, indeed this was a bug.

lib/handlers/ortc/edgeUtils.js
```diff
@@ -75,6 +75,11 @@ exports.mangleRtpParameters = function(rtpParameters)
 			delete codec.channels;
 		}
 
+		if (codec.mimeType && !codec.name)
+		{
+			codec.name = codec.mimeType.split('/')[1];
```
ibc, line 80: We don't use `codec.name` but just `codec.mimeType` so this is not needed.

ibc:
Thanks! I've made minor modifications and released it in 3.2.3 (already in NPM):

https://github.com/versatica/mediasoup-client/commits/v3

lib/handlers/ortc/edgeUtils.js
milen-yordanov: Edge uses codec.name. If codec.name is not specified it returns InvalidAccessError.
See: https://draft.ortc.org/#rtcrtpcodecparameters*

RTCRtpCodecParameters.name seems to be required.

ibc:
You are completely right, my fault. Indeed adding `codec.name` is required in `edgeUtils.mangleRtpParameters()`, otherwise `this._handler.receive({ id, kind, rtpParameters });` in `transport.consume()` will fail.

Let me add it.

lib/handlers/ortc/edgeUtils.js
ibc: You are completely right, my fault. Indeed adding `codec.name` is required in `edgeUtils.mangleRtpParameters()`, otherwise `this._handler.receive({ id, kind, rtpParameters });` in `transport.consume()` will fail.

Let me add it.

ibc: I've pushed to v3 branch. Can you confirm it works before I release 3.2.4, please?

Thanks a lot.

milen-yordanov: The v3 branch now works. Thanks.

ibc: Great. Released in mediasoup-client 3.2.4.

Thanks!

PR #92, opened 2019-10-17
Fix an error of producing with `react-native-webrtc`

js-seo:
When creating an offer(`rtcPeerConnection.createOffer()`),
the default argument is `{ offerToReceiveAudio: true, offerToReceiveVideo: true }`.

https://github.com/react-native-webrtc/react-native-webrtc/blob/master/RTCPeerConnection.js#L121

and using this causes an error like below on setting remote description.

```
{
  name: "SetRemoteDescriptionFailed",
  message: ""Failed to set remote answer sdp: The order of m-lines in answer doesn't match order in offer. Rejecting answer.""
}
```

From what I understand, mediasoup does not allow a sending transport to receive media at the same time.
So it seems that offering to receive at the same time was not accepted by mediasoup server.

### DIFF
--- a/lib/handlers/ReactNative.js
+++ b/lib/handlers/ReactNative.js
@@ -171,7 +171,10 @@ class SendHandler extends Handler
 		this._stream.addTrack(track);
 		this._pc.addStream(this._stream);
 
-		let offer = await this._pc.createOffer();
+		let offer = await this._pc.createOffer({
+			offerToReceiveAudio : false,
+			offerToReceiveVideo : false
+		});
 		let localSdpObject = sdpTransform.parse(offer.sdp);
 		let offerMediaObject;
 		const sendingRtpParameters =


parkjurung:
It seems very reasonable.

zhangzhx666:
I've had the same problem When I call enbaleMic().I call  'createOffer({offerToReceiveAudio: true, offerToReceiveVideo: false})',it's ok.

ibc:
Thanks a lot, @js-seo. I've fixed it in https://github.com/versatica/mediasoup-client/commit/95c5de4cd4fe16524606a0835ffeaa6cc8c83a1c. Note that I've also added `offerToReceiveXxxxx: false` in more `createOffer()` calls, such as when creating a DataChannel (`DataProducer` or `DataConsumer` in mediasoup).

It's released in mediasoup-client 3.2.7.

@zhangzhx666: `enableMic()` is not part of mediasoup-client but is just part of the  mediasoup-demo project. Please don't assume they both are the same thing.

ibc:
I've announced the new version in the mediasoup Discourse Group: https://mediasoup.discourse.group/t/mediasoup-client-3-2-7-released-with-important-fix-for-react-native/329

ibc:
BTW @saghul, having those `createOffer()` [default options](https://github.com/react-native-webrtc/react-native-webrtc/blob/master/RTCPeerConnection.js#L45) is like something from 1996, am I wrong?

saghul:
They are hace at this point. But people keep making the mistake of not putting them so we left them in.

ibc:
People do not put them because their default value is `false` and not `true` as react-native-webrtc assumes. Ok, not extremely clear in the current WebRTC 1.0 spec, but that's the behaviour of any browser nowadays:

```js
const pc = new RTCPeerConnection();
const offer = await pc.createOffer()

offer.sdp

// =>

"v=0
o=- 3160023335042736945 2 IN IP4 127.0.0.1
s=-
t=0 0
a=msid-semantic: WMS
"
```

saghul:
Can you open an issue on react-native-webrtc? I'll be hapy to remove them for next release.

ibc:
[Done](https://github.com/react-native-webrtc/react-native-webrtc/issues/712)

PR #111, opened 2020-02-20
Update tsconfig not to polyfill async/await

leaysgur:
Current settings in `tsconfig.target` is `es6`.
This convert this

```js
(async () => {
  await new Promise(r => setTimeout(r, 1000));
}())
```

into this.

```js
var __awaiter = (this && this.__awaiter) || function (thisArg, _arguments, P, generator) {
    function adopt(value) { return value instanceof P ? value : new P(function (resolve) { resolve(value); }); }
    return new (P || (P = Promise))(function (resolve, reject) {
        function fulfilled(value) { try { step(generator.next(value)); } catch (e) { reject(e); } }
        function rejected(value) { try { step(generator["throw"](value)); } catch (e) { reject(e); } }
        function step(result) { result.done ? resolve(result.value) : adopt(result.value).then(fulfilled, rejected); }
        step((generator = generator.apply(thisArg, _arguments || [])).next());
    });
};
(() => __awaiter(void 0, void 0, void 0, function* () {
    yield new Promise(r => setTimeout(r, 1000));
}))();
```

Since `mediasoup-client` requires developers to use bundler, polyfill also should be responsible to developers, I think.

---

1 more question.
Why are you adding built `/lib/**/*` files under `git` and publish raw `/src/**/*.ts` files(and others including test files) to npm?

### DIFF
--- a/lib/Consumer.js
+++ b/lib/Consumer.js
@@ -1,13 +1,4 @@
 "use strict";
-var __awaiter = (this && this.__awaiter) || function (thisArg, _arguments, P, generator) {
-    function adopt(value) { return value instanceof P ? value : new P(function (resolve) { resolve(value); }); }
-    return new (P || (P = Promise))(function (resolve, reject) {
-        function fulfilled(value) { try { step(generator.next(value)); } catch (e) { reject(e); } }
-        function rejected(value) { try { step(generator["throw"](value)); } catch (e) { reject(e); } }
-        function step(result) { result.done ? resolve(result.value) : adopt(result.value).then(fulfilled, rejected); }
-        step((generator = generator.apply(thisArg, _arguments || [])).next());
-    });
-};
 Object.defineProperty(exports, "__esModule", { value: true });
 const Logger_1 = require("./Logger");
 const EnhancedEventEmitter_1 = require("./EnhancedEventEmitter");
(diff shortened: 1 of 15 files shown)


ibc:
> Since mediasoup-client requires developers to use bundler, polyfill also should be responsible to developers, I think.

This is not entirely true. mediasoup-client works also in Node where no bundling is needed at all.

> Why are you adding built /lib/**/* files under git and publish raw /src/**/*.ts files(and others including test files) to npm?

Because we want it to work if the user installs it via `npm install mediasoup` or by importing mediasoup-client from a local repository.

Coming back to the subject of this PR:

I'm fine with generating `es2020` but we must verify that the generated output still works in Node versions X.

ibc:
I've done some checks and applied your commit manually and pushed to "v3" branch. Will release a new mediasoup-client version soon.

Thanks

ibc:
3.3.17 released

leaysgur:
Oh, I overlooked that this module supports Node too.

> Because we want it to work if the user installs it via npm install mediasoup or by importing mediasoup-client from a local repository

FYI, you may use `files` field in `package.json` for that purpose.
But for local repository, you need to build them once.

Anyway, thanks though!

ibc:
> FYI, you may use `files` field in package.json for that purpose.

May you explain it a bit?

BTW: I want to be able to add `mediasoup-client` with GIT URL as a dependency in a `package.json` and use it without having to enter into `node_modules/mediasoup-client/`, running `npm install` (to install dev dependencies) and running `npm run postinstall`.

leaysgur:
By using [`files`](https://docs.npmjs.com/files/package.json#files) field, we can specify which files to be uploaded to `npm` registry.(= white list, use `.npmignore` for black list)

In general, users don't need to download(via `npm install`) raw source(`.ts`), tests, build scripts and so on, specify only built results(`.js` files and maybe `.d.ts`) in `files` field is reasonable.

But according to your case, it seems built result(`.js` files) must be under the Git. This field is useless. 😅

ibc:
At the end we need `.ts`, `.js` and `.d.ts` in Git, and we need `.js` and `.t.ds` in NPM. So we may filter `.ts` in NPM. Can be considered although I there are just a few `. ts` files...

PR #114, opened 2020-03-24
fix: parameters not being added to the receiver

ethand91:
# The problem

The producing uses the parameters set in the server side config, however the receiver does not and instead uses the defaults generated from createOffer().

### Example

mediasoup server side config
```javascript
mediaCodecs: [
  {
    kind: 'audio',
    mimeType: 'audio/opus',
    clockRate: 48000,
    channels: 2,
    parameters: {
      stereo: 1,
      'sprop-stereo': 1
    }
  }
]
```

Using the above configuration the producer produces an audio stream using stereo,
however when receiving the stream the audio codec parameters are the default which means the audio will by played via mono.

This pull request fixes this by adding the parameters to the producer's consumableRtpParameters and adds them to the consumer's rtpCapabilities, enabling the consumer to play the audio via stereo.

If there is a better way or any issues with this implementation please let me know.


ibc:
This is a good point. However this is already solved and **working** in a different way in mediasoup and mediasoup-client (and libmediasoupclient):

* Router RTP capabilities represent what the server can receive (which specific codec configurations) and/or what is willing to receive (certain codec features/parameters).
* However those parameters are just a hint for the client.
* If the client wishes to enable OPUS stereo, you don't need to force it for everyone in the Router. Just call `transport.produce()` in mediasoup-client with `opusStereo: true` in `codecOptions`:

https://mediasoup.org/documentation/v3/mediasoup-client/api/#ProducerOptions

BTW this is done in the demo:

- https://v3demo.mediasoup.org/?externalVideo=true&roomId=foo&consume=false
- https://v3demo.mediasoup.org/?externalVideo=true&roomId=foo&produce=false

PR #118, opened 2020-05-28
Fakehandler

jmillan:


src/handlers/FakeHandler.ts
```diff
@@ -0,0 +1,362 @@
+import { Logger } from '../Logger';
+// @ts-ignore
```
ibc: Not needed. It exposes TS types.

ibc (approved):
LGTM. Just a comment

PR #136, opened 2020-10-01
Add Observer API

jmillan:


jmillan:
NOTE: Doc is missing. On it.

src/Consumer.ts
```diff
@@ -246,6 +269,9 @@ export class Consumer extends EnhancedEventEmitter
 
 		this._paused = false;
 		this._track.enabled = true;
+
+		// Emit observer event.
+		this._observer.safeEmit('resume');
```
ibc, line 275: What about `trackended` event?

src/Device.ts
```diff
@@ -331,6 +334,16 @@ export class Device
 		return this._sctpCapabilities!;
 	}
 
+	/**
+	 * Observer.
+	 *
+	 * @emits newtransport - (transport: Transport)
```
ibc, line 340: There is no  `Transport` class but `SendTransport` and `RecvTransport `.

jmillan: correct

ibc: Forget about this, my fault.

ibc (approved):
LGTM, just a minor thing commented

src/Consumer.ts
jmillan: done

Issue #145, opened 2020-12-06
Firefox handler doesn't respect RtpSendParameter.encodings if only 1 encoding is specified 


andrewlin12:
## Bug Report
Firefox does not respect `encodings` if only a single encoding specification is passed to `Transport.produce`.

### Your environment

- Operating system: MacOS
- Browser version: Firefox 83
- mediasoup-client version: 3.6.21

### Issue description
In the Firefox60 handler, it looks like `encodings` is ignored if length is 1: https://github.com/versatica/mediasoup-client/blob/v3/src/handlers/Firefox60.ts#L307. I tested a modification to initialize `reverseEncodings` to `encodings` and Firefox started respecting the parameters.

ibc:
BTW note that Firefox is not yet spec compliant regarding encodings:

https://github.com/versatica/mediasoup-client/blob/v3/src/handlers/Firefox60.ts#L341

However we do the trick there. If true that we should consider single encoding case and respect them. It should be done here:

https://github.com/versatica/mediasoup-client/blob/v3/src/handlers/Firefox60.ts#L317

If there are encodings but just one we should assign them to reverseEncodings there.

Can you do a PR? Otherwise I'll just fix it eventually when I get time.

andrewlin12:
FYI, I posted a possible fix here: https://github.com/versatica/mediasoup-client/pull/146

ibc:
Yup, thanks. Will review tomorrow.

PR #146, opened 2020-12-06
Change Firefox60 handler to respect single RtpEncodingParameter

andrewlin12:
### Description
This fix addresses #145  - Firefox 60 was only respecting `RtpSendParameters.encodings` if more than one specified. 

### Change
In `Firefox60.ts`, always clone `encodings` if it is non-empty, and only reverse it if there is more than one element. 

One interesting note - if you try to set `rid` on the solo encoding, Firefox doesn't seem to respect it.


lib/handlers/Firefox60.js
```diff
@@ -184,13 +184,16 @@ class Firefox60 extends HandlerInterface_1.HandlerInterface {
         this._assertSendDirection();
         logger.debug('send() [kind:%s, track.id:%s]', track.kind, track.id);
         let reverseEncodings;
-        if (encodings && encodings.length > 1) {
-            encodings.forEach((encoding, idx) => {
-                encoding.rid = `r${idx}`;
```
andrewlin12, line 189: Curious if the mutation of the original `encodings` here is intentional?

src/handlers/Firefox60.ts
```diff
@@ -304,16 +304,20 @@ export class Firefox60 extends HandlerInterface
...
 		let reverseEncodings;
 
-		if (encodings && encodings.length > 1)
+		if (encodings && encodings.length > 0)
 		{
-			encodings.forEach((encoding, idx) =>
+			reverseEncodings = utils.clone(encodings, []);
+			if (encodings.length > 1)
 			{
-				encoding.rid = `r${idx}`;
```
andrewlin12, line 311: Is the mutation of the passed-in `encodings` is intentional?

ibc: Yes, because if multiple `encodings` are given then `rid` is mandatory. However we do it internally in mediasoup-client rather than requiring the user to do it.

ibc (approved):

ibc:
Released in 3.6.22, thanks.

ibc:
Just a thing. This PR was wrong aqnd it failed by just running the mediasoup demo by enabling webcam with simulcast in Firefox.

https://github.com/versatica/mediasoup/issues/486

PR #149, opened 2020-12-27
Reuse closed m= sections in Firefox

snnz:
This PR handles m= sections reuse in Firefox by synchronizing RemoteSdp with the local description regardless of where the new transceiver's section is placed by the browser.


ibc:
Thanks. AFAIR there is a reason for me to not do this that way in the past, but I cannot recall which one now. I have to take a deeper look to this. IMHO if this works it must be done for all handlers and not just for Firefox.

src/handlers/sdp/RemoteSdp.ts
```diff
@@ -442,6 +447,45 @@ export class RemoteSdp
 		}
 	}
 
+	_syncMediaWithLocalSdp(localSdpMedia: any, newMediaSection: MediaSection): void
+	{
+		if (!this._firstMid)
+			this._firstMid = newMediaSection.mid;
+
+		// Append new section to the existing vector and the SDP object
```
jmillan: Dot after en of comments.

src/handlers/sdp/RemoteSdp.ts
```diff
@@ -442,6 +447,45 @@ export class RemoteSdp
...
+		let idx = this._mediaSections.length;
+		this._mediaSections.push(newMediaSection);
+		this._sdpObject.media.push(newMediaSection.getObject());
+
+		// Add it to the map
```
jmillan: Same for all the added comments.

src/handlers/sdp/RemoteSdp.ts
```diff
@@ -442,6 +447,45 @@ export class RemoteSdp
...
+			{
+				this._mediaSections.push(mediaSections[i]);
+				this._sdpObject.media.push(media[i]);
+			}
+		}
+
+		// Recreate map
+		this._midToIndex.clear();
+		for (idx = 0; idx < this._mediaSections.length; ++idx)
+			this._midToIndex.set(this._mediaSections[idx].mid, idx);
```
jmillan: braces missing.

src/handlers/sdp/RemoteSdp.ts
snnz: Fixed.

src/handlers/sdp/RemoteSdp.ts
snnz: Periods added to the comments.

src/handlers/sdp/RemoteSdp.ts
snnz: Braces added.

ibc:
I will come back to this PR next week.

ibc (commented):
Can you please explain in detail what this PR does and what it fixes? And why does it allow closing the first media section? That will break/close the ICE connection globally.

snnz:
It does not allow closing the first media section, since it uses regular `_remoteSdp.closeMediaSection` function, which checks for the first section.

The purpose of this PR was to allow the Firefox handler to reuse closed media section on sending, as other handlers do. Currently, when some producers are closed, and new ones created later, it just keep adding new media sections to SDP, so in case a lot of producers are created and closed in the lifetime of the connection, SDP can grow quite a lot. The problem with Firefox is that when adding new transceiver, it uses not first available closed section, but the one with media type (audio/video) matching the type of the new transceiver. Thus, the method of finding next media section in the internal array (which seaches for the first closed section) does not work with Firefox. This is why closing any sections was disabled the Firefox handler.
But still Firefox reuses closed media section quite normally. It just that some other method should be used to determine where the new transceiver is landed in the SDP.

I used the fork from which the PR is created in the real system for quite a while without any problems. But recently Firefox started to fail on the mismatching media types in the _receiving_ connection. I took a look at the commits in the upstream, saw the one that fixes that problem, merged the changes into the fork, and the PR was bumped up. After all these years, I almost forgot about it. :)
And actually, I don't remember why rather complex way of dealing with the problem was chosen, with recreating and rearranging entire media sections array. It seems that after the index of the new trainsceiver is found in the local SDP, the `mid` of the previous section to reuse can be simple obtained by this index, and then used in the same way as in the other handlers. The only addition necessary for RemoteSdp class is a public function that returns media section by index (providing that the index within the boundaries of the array).
Please take a look at the new version, it should be much сlearer what it does.

ibc:
Thanks. I will check in next week.

PR #151, opened 2021-01-02
Reverse simulcast encondings order in Firefox before sending them to the server

snnz:


ibc:
Thanks a lot. I've fixed this in https://github.com/versatica/mediasoup-client/commit/3f2f7600c2f3cefb43e102392b17312e6bf9a746, which is similar to your PR but a bit better (not just if VP8 or H264, although it's finally the same).

PR #154, opened 2021-01-17
Fixed incompatibility with newer TypeScript versions

Minding000:
Removed `async` keyword, since it is not allowed with the `abstract` keyword (Error **TS1243**).

The implementation is still clear from the method signature (`Promise` as return type).


ibc:
Thanks, released in 3.6.27.

PR #161, opened 2021-03-03
Fix ScalabilityMode issue

holtwick:
This version works with Vite.js


package.json
```diff
@@ -39,24 +39,24 @@
...
+    "eslint-plugin-jest": "^24.1.5",
     "fake-mediastreamtrack": "^1.1.6",
     "jest": "^26.6.3",
     "jest-tobetype": "^1.2.3",
     "open-cli": "^6.0.1",
     "tsc-watch": "^4.2.9",
-    "typescript": "~3.7.5",
-    "uuid": "^3.4.0"
+    "typescript": "~4.2.2",
+    "uuid": "^8.3.2"
```
ibc: Please let's not update deps (unrelated to this PR).

- Also AFAIR `uuid` 8.X.X has some issues in React-Native.
- Also, there are reasons for we to stay in typescript 3.X.X (otherwise Angular users will complain).

I'll take the valid changes and apply them into my ongoing PR.

holtwick: Sure, will reduce to the basic requirements.

ibc: AFAIS `export type` does not work with typescript 3.X.X.

holtwick: Should work since TS 3.8 https://www.typescriptlang.org/docs/handbook/modules.html#importing-types

tsconfig.json
```diff
@@ -7,6 +7,7 @@
 		"module": "commonjs",
 		"moduleResolution": "node",
 		"esModuleInterop": true,
+    "isolatedModules": true,
```
ibc: Please use TAB here instead of spaces. Indentation is broken otherwise.

ibc (changes requested):

package.json
ibc: Great

ibc (approved):
Thanks, let's try it.

tsconfig.json
holtwick: Maybe a Prettier action brings everybody easily on the same page ;) https://github.com/marketplace/actions/prettier-action

ibc: I know prettier, but prefer to not use it in mediasoup-client yet, perhaps in the future.

PR #165, opened 2021-05-01
iOS WKWebView support

holtwick:
On iOS only look for OS version >= 14.3, which is the one where getUserMedia was released

In custom iOS apps the user agent string is not recognized by Bowser. Instead, check for OS version.

It would be great, if the internal check could be fully overridden if the user is sure it will work.

### DIFF
--- a/src/Device.ts
+++ b/src/Device.ts
@@ -80,6 +80,13 @@ export function detectDevice(): BuiltinHandlerName | undefined
 		const browser = Bowser.getParser(ua);
 		const engine = browser.getEngine();
 
+    const browserInfo = Bowser.parse(window.navigator.userAgent)
+    const [major, minor, patch] = browserInfo?.os?.version?.split(".") || []
+    const isIOSWebKit =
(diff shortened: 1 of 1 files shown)


src/Device.ts
```diff
@@ -80,6 +80,13 @@ export function detectDevice(): BuiltinHandlerName | undefined
 		const browser = Bowser.getParser(ua);
 		const engine = browser.getEngine();
 
+    const browserInfo = Bowser.parse(window.navigator.userAgent)
```
ibc, line 83: There is no window in Node and mediasoup-client works in Node. You are using spaces instead of tab, so breaking `npm run lint` task.

Please read the docs properly. You can choose DeviceName via API.

holtwick: Sorry, you are right: https://mediasoup.org/documentation/v3/mediasoup-client/api/#Device-dictionaries

I didn't find that solution before.

holtwick:
@ibc  This can be solved as you described before. Thanks for giving the hint.

PR #166, opened 2021-06-03
allow override of dtls role on send/receive

artushin:
I'm not sure what the benefit of passing the dtls role of `server` is on producers, but having the ability to override to `client` so that the server can determine crypto algorithms is useful. Attached is a proposal for this override.


ibc:
I'm afraid this PR is wrong. DTLS role is not a Producer/Consumer setting but a Transport setting. There is no sense in passing `localDtlsRole` in `produce()` or `consume()` calls.

artushin:
No problem, mostly this PR was just to see if you were ok with the override itself. I can update it for better design.

Setting it via transport options was what I initially looked at doing but _setupTransport happens off the calls to send/receive and that's where the following occurs in the handlers:

```
		// Get our local DTLS parameters.
		const dtlsParameters =
			sdpCommonUtils.extractDtlsParameters({ sdpObject: localSdpObject });

		// Set our DTLS role.
		dtlsParameters.role = localDtlsRole;
```

I assumed this functionality is handler specific but looking again it doesn't seem to be. It seems like (in RemoteSdp) you should be able to just wrap

```
this._dtlsParameters!.role = role;
```

in a conditional

```
if (!this._dtlsParameters!.role) {
  this._dtlsParameters!.role = role;
}
```

and then you just pass the role as part of the transport options and the _setupTransport doesn't override it.

You could then just remove the 
```
		// Set our DTLS role.
		dtlsParameters.role = localDtlsRole;
```

from _setupTransport in the handlers.

artushin:
@ibc Any thoughts on my previous comment?

ibc:
I'm afraid I won't be able to take a look at this until next week. It's in my radar.

artushin:
Thanks!

artushin:
@ibc Any updates on when this might be able to go in?

ibc:
Sorry for the late response. This feature requires a different API than the one you did in this PR and I need some time before I get involved into it (too busy for now).

artushin:
Ok, let me know if you have an idea about the API and I might be able to do the proposed implementation.

artushin:
@ibc Any update on this?

ibc:
There is another issue with more rationale about this and this feature it's gonna be added, I promise. It's just that we are kinda off for long weeks.

ibc:
This is implemented in medisoup-client 3.6.44, see rationale in the announcement: https://mediasoup.discourse.group/t/medisoup-client-3-6-44-released-with-ability-to-force-local-dtls-role/3571

PR #173, opened 2021-07-14
Allow setting RtpEncodingParameters for individual encodings

t-mullen:
## Description

Adds an optional second parameter to `Producer::setRtpEncodingParameters` that allows applications to specify the index of the encoding to modify. Not providing this second parameter will modify all encodings just like before.

## Usecases
I have a few uses for this - just some examples:
1) Disabling lower-indexed encodings (similar to `setMaxSpatialLayer` but reversed).
2) Modifying `scaleResolutionDownBy` depending on the current source resolution of the track.


nazar-pc:
In Firefox encodings are in the reverse order comparing to other browsers, this doesn't seem to account for that.

t-mullen:
After some discussion elsewhere, going to reopen with an a slightly different API. #174

PR #174, opened 2021-07-14
Allow setting RtpEncodingParameters for individual encodings

t-mullen:
## Description

Changes `Producer::setRtpEncodingParameters` to also accept an array of RTCRtpEncodingParameters, for setting different parameters for each encoding. Passing a single RTCRtpEncodingParameters object will work as before.

## Usecases
I have a few uses for this - just some examples:
1) Disabling lower-indexed encodings (similar to `setMaxSpatialLayer` but reversed).
2) Modifying `scaleResolutionDownBy` depending on the current source resolution of the track.


src/handlers/Chrome67.ts
```diff
@@ -541,10 +541,20 @@ export class Chrome67 extends HandlerInterface
 
 		const parameters = rtpSender.getParameters();
 
-		parameters.encodings.forEach((encoding: RTCRtpEncodingParameters, idx: number) =>
+		if (Array.isArray(params)) 
 		{
-			parameters.encodings[idx] = { ...encoding, ...params };
-		});
+			parameters.encodings.forEach((encoding: RTCRtpEncodingParameters, idx: number) =>
+			{
```
ibc, line 547: This is wrong. Old matching properties override the new ones.

t-mullen: This is the way property assignment was done before. It seems to be correct:
```js
const oldObj = { x: 1 }
const newObj = { x: 2 }
{ ...oldObj, ...newObj } // { x: 2 }
```

It transpiles to:
```js
parameters.encodings[idx] = Object.assign(Object.assign({}, encoding), params[idx]);
```
Which also seems right.

ibc: I'm in mobile and could be wrong. I'll review in next days. Thanks

jmillan (approved):
LGTM

ibc:
I want to re-review this, sorry for the delay, off for many weeks.

PR #175, opened 2021-07-15
Add missing "active" property to RtpEncodingParameters type

t-mullen:
### Description
Adds the missing "active" property to `RtpEncodingParameters` type which is otherwise a superset of [RTCRtpEncodingParameters](https://developer.mozilla.org/en-US/docs/Web/API/RTCRtpEncodingParameters).

### Usecase
Calling `Producer::setRtpEncodingParameters` to activate/deactivate encodings.


ibc:
Actually the public API of mediasoup client doesn't allow setting `active` field of encoding on purpose. This is not just about exposing `active` in TS types. It's that literally the `active` field is ignored if present.

PR #178, opened 2021-08-12
Add support for concurrent handler operations in Chrome74

ggarber:
### Motivation

Right now every time there is a change in the number of consumers or producers that requires an additional offer/answer renegotiation in the PeerConnection and those operations are serialized (executed sequentially) having to wait for the finish of the previous one to start the next one.

Because of that joining a room with many participants is a very slow process right now.  For example joining a room with 80 participants (160 consumers) takes 30 seconds in my laptop.    To make things worse it looks like renegotiations take longer the more consumers you have so having 100 participants can take more than 60 seconds some times.

All those operations can be done in parallel adding multiple consumers/producers in a single offer/answer negotiation.  With this change the same test joining a room with 80 participants takes **3 seconds instead of 30 seconds** in my laptop.

### Description

Today when a Transport needs to do a renegotiation (f.e. when creating or destroying consumers) it makes sure that there isn't other renegotation in progress with an async queue in the Transport.  So the Handler doesn't need to be aware of it, it can just assume that renegotiations will never overlap.

With this PR when a Transport needs to do a renegotiation (f.e. when creating or destroying consumers) it just requests it from the Handler without passing through the async queue (for the handlers that supports it that right now is only implemented in Chrome74 handler).   So the Handler needs to be able to handle renegotiation requests received before finishing the previous ones.

The way that is implemented is with a flag inside the handler to mark when there is renegotiation in progress (the `_negotiationInProgress` promise) and a queue to store the information for all the renegotiations received while that renegotiation is in progress and that will be executed once the previous one is finished (the `_negotiationsQueued` list).

So if there is no renegotiation in progress the handler will behave like today, just starting a renegotiation as soon as possible.   But if there is a renegotiation in progress it will be queued until the previous is finished and then start a new one to renegotiate the parameters for ALL the renegotiation requests that happened in the middle.     Usually if you create 100 consumers for example that means that one renegotiation will be triggered for the first one, then 99 received later will be queued and when the first one finishes there will be ONE new renegotiation to apply the parameters of those 99 queued.

The implementation is done creating helper functions to encapsulate the renegotiations (`renegotiateSend` and `renegotiateReceive`) so that the renegotiations are in a single place.   Those helper functions receive some callbacks to be able to mangle or read parameters from the SDPs during the renegotiation progress.  In case of `renegotiateSend` there are callbacks to do something after creating the offer and also after SetLocalDescription.  In case of `renegotiateReceive` there are callbacks to do something after creating the asnser.

### Test Plan

It was tested as part of our application:
* Manual mode: Joining/leaving rooms with multiple participants and checking audio and video works properly.
* Automated tests: Running our test suite + stress tests creating 100 producers in parallel, 100 consumers in parallel and 100 consumer start/stop in parallel.


ibc:
This is super interesting plus a MUST-HAVE. Not sure how good this is in August before some deserved or undeserved holidays, but definitely will take a look to (`ETA = NaN`)  it and discuss it with you in place.

jmillan:
The feature makes a lot of sense @ggarber  :+1:

fippo:
I'd suggest testing in chrome 93 which has all the fixes from https://bugs.chromium.org/p/webrtc/issues/detail?id=12840 -- the perf was increased quite a bit.

ggarber:
I have submitted the whole changes needed for Chrome74.   The only important changes are in the Chrome74.ts file where everything is shuffled around a bit.

It is a big and tricky/risky change but a must at least for use cases with many consumers in parallel (typical videoconference room with 50+ participants).   

Maybe there is a better way to implement it but i couldn't find it, at least without modifying the public APIs.   Happy to consider other alternatives if needed.

Find in the PR summary the description of how I have tested it, if there is a way to easily test it in other apps like mediasoup-demo it would be great.   In the mean time we probably apply this change locally and use it with real users as it is a blocker for us.

Thank you very much and when you have some time let me know what you think or how to proceed/abandon with this change.

jmillan:
Hi @ggarber,

> This PR implementes that idea for Chrome74 handler (producers, consumers and datachannels) by using a promise to detect when there is a negotiation in progress and wait for it + using a queue of callbacks so that we can give a chance to modify the SDPs if needed for all the pending negotiations once they are done in a single renegotiation.

Before going for a deeper review, could you please explain a bit in more detail what the PR does? 

Thanks

ggarber:
> Hi @ggarber,
> 
> > This PR implementes that idea for Chrome74 handler (producers, consumers and datachannels) by using a promise to detect when there is a negotiation in progress and wait for it + using a queue of callbacks so that we can give a chance to modify the SDPs if needed for all the pending negotiations once they are done in a single renegotiation.
> 
> Before going for a deeper review, could you please explain a bit in more detail what the PR does?
> 
> Thanks

Hi @jmillan, tried to explain it better in the "Description" section of the PR summary.   Let me know if that doesn't help.

I'm also open to a quick call to talk about it or review it together if you want.

ibc:
The description is now perfect. So just please let's us review when possible.

jmillan (approved):
LGTM, nice

Datachannel is also tested right, @ggarber?

ggarber:
@jmillan We have seen 2 issues in production with this change (unfrequent and only affecting really big rooms) so probably you should wait to land it until we figure out what is the exact issue.

jmillan:
> @jmillan We have seen 2 issues in production with this change (unfrequent and only affecting really big rooms) so probably you should wait to land it until we figure out what is the exact issue.

Thanks for notifying @ggarber. We'll also re-review the code.

jmillan:
@ggarber, did you get to the root issue? Is this code working as expected? Thanks.

jmillan:
We are working on another approach to reach concurrency. It's WIP.

jmillan:
@ggarber, here the PR enabling concurrent consumer creation in all handlers. https://github.com/versatica/mediasoup-client/pull/191

ibc:
Let's close this PR in favour of #191 (which is in review process).

PR #180, opened 2021-09-21
Add pause/resume receiving support

ggarber:
Right now when pausing a consumer we are only disabling the playback/rendering of the corresponding track (track.enabled = false) but the underlaying transceiver is still decoding and processing the audio/ video samples.   This has at least two consequences:
- Audio/Video are still being decoded if they are received.
- Even if audio samples are not received the audio pipeline keeps trying to generate audio samples because it thinks that audio packets are missing.

Both things have implications in CPU usage and the one that I'm more worried about is the second one.   The use case is  having a big room (f.e. 100 participants) and most of them are muted.   Without this change one client in the room has 100 active transceivers (even if the track is disabled) and those transceivers are trying to generate fake audio samples.  Look at the capture from webrtc-internals in the receiver side while the remote producer is muted/paused (at 12:03 it was unmuted for 20 secs and rest of the time it was muted):

Screenshot 2021-09-20 at 12 06 45

This change avoids that extra processing disabling the corresponding transceiver when pausing a consumer.   This is implemented by setting direction = inactive instead of direction = readonly and generating a renegotiation without any modification in the remote description.

It has only been implemented for the latest version of the handlers (Chrome74, Firefox60 and Safari12) where support for transceivers is available.  It could be implemented for all of them with SDP mangling but i'm not sure if that's needed given that this is a performance optimization.
 
 It is possible that the audio decoding is still happening even with the transceiver disabled if audio samples are received (https://bugs.chromium.org/p/webrtc/issues/detail?id=11013&) but that's a bug and not an issue in case the producer is muted and not sending audio samples as far as I can tell.


ibc:
What's the benefit of this?

nazar-pc:
Prevent browser from decoding video I assume?

ibc:
> Prevent browser from decoding video I assume?

This is what I want to confirm :)

But, if true, wouldn't track.enabled=false do the same? Just wondering.

ggarber:
Sorry I created the PR and then had to take care of the kids before I had time to write a proper description so I left it as draft :)

ibc:
Ok ping here please when done

ibc:
I think it's not needed to do this also in old handlers. I'll review tomorrow.

src/handlers/Chrome55.ts
```diff
@@ -653,6 +653,20 @@ export class Chrome55 extends HandlerInterface
 
 		await this._pc.setLocalDescription(answer);
 	}
+	
+	async pauseReceiving(
+		// eslint-disable-next-line @typescript-eslint/no-unused-vars
+		localId: string): Promise<void>
+	{
+		// Unimplemented
```
ibc:
```suggestion
		// Unimplemented.
```

src/handlers/Chrome55.ts
```diff
@@ -653,6 +653,20 @@ export class Chrome55 extends HandlerInterface
...
+	}
+
+	async resumeReceiving(
+		// eslint-disable-next-line @typescript-eslint/no-unused-vars
+		localId: string): Promise<void>
+	{
+		// Unimplemented
```
ibc:
```suggestion
		// Unimplemented.
```

src/handlers/Chrome67.ts
```diff
@@ -733,6 +733,20 @@ export class Chrome67 extends HandlerInterface
 		await this._pc.setLocalDescription(answer);
 	}
 
+	async pauseReceiving(
+		// eslint-disable-next-line @typescript-eslint/no-unused-vars
+		localId: string): Promise<void>
+	{
+		// Unimplemented
```
ibc:
```suggestion
		// Unimplemented.
```

src/handlers/Chrome67.ts
```diff
@@ -733,6 +733,20 @@ export class Chrome67 extends HandlerInterface
...
+	}
+
+	async resumeReceiving(
+		// eslint-disable-next-line @typescript-eslint/no-unused-vars
+		localId: string): Promise<void>
+	{
+		// Unimplemented
```
ibc:
```suggestion
		// Unimplemented.
```

src/handlers/Chrome70.ts
```diff
@@ -735,6 +735,20 @@ export class Chrome70 extends HandlerInterface
 		await this._pc.setLocalDescription(answer);
 	}
 
+	async pauseReceiving(
+		// eslint-disable-next-line @typescript-eslint/no-unused-vars
+		localId: string): Promise<void>
+	{
+		// Unimplemented
```
ibc:
```suggestion
		// Unimplemented.
```

src/handlers/Chrome70.ts
```diff
@@ -735,6 +735,20 @@ export class Chrome70 extends HandlerInterface
...
+	}
+
+	async resumeReceiving(
+		// eslint-disable-next-line @typescript-eslint/no-unused-vars
+		localId: string): Promise<void>
+	{
+		// Unimplemented
```
ibc:
```suggestion
		// Unimplemented.
```

src/handlers/Chrome74.ts
```diff
@@ -721,6 +721,66 @@ export class Chrome74 extends HandlerInterface
...
+		logger.debug('pauseReceiving() [localId:%s]', localId);
+
+		const transceiver = this._mapMidTransceiver.get(localId);
+
+		if (!transceiver)
+			throw new Error('associated RTCRtpTransceiver not found');
+
+		transceiver.direction = 'inactive';
+		
+		const offer = { type: 'offer', sdp: this._remoteSdp!.getSdp() };
```
ibc, line 737: Are you sure this works? I'm pretty sure that `this._remoteSdp!.getSdp()` is gonna generate a SDP with `a=recvonly`. Wouldn't it invalidate previous `transceiver.direction = 'inactive'`?

src/handlers/Edge11.ts
```diff
@@ -424,6 +424,20 @@ export class Edge11 extends HandlerInterface
 		}
 	}
 
+	async pauseReceiving(
+		// eslint-disable-next-line @typescript-eslint/no-unused-vars
+		localId: string): Promise<void>
+	{
+		// Unimplemented
```
ibc:
```suggestion
		// Unimplemented.
```

src/handlers/Edge11.ts
```diff
@@ -424,6 +424,20 @@ export class Edge11 extends HandlerInterface
...
+	}
+
+	async resumeReceiving(
+		// eslint-disable-next-line @typescript-eslint/no-unused-vars
+		localId: string): Promise<void>
+	{
+		// Unimplemented
```
ibc:
```suggestion
		// Unimplemented.
```

src/handlers/FakeHandler.ts
```diff
@@ -348,6 +348,20 @@ export class FakeHandler extends HandlerInterface
 		this._tracks.delete(Number(localId));
 	}
 
+	async pauseReceiving(
+		// eslint-disable-next-line @typescript-eslint/no-unused-vars
+		localId: string): Promise<void>
+	{
+		// Unimplemented
```
ibc:
```suggestion
		// Unimplemented.
```

src/handlers/FakeHandler.ts
```diff
@@ -348,6 +348,20 @@ export class FakeHandler extends HandlerInterface
...
+	}
+
+	async resumeReceiving(
+		// eslint-disable-next-line @typescript-eslint/no-unused-vars
+		localId: string): Promise<void>
+	{
+		// Unimplemented
```
ibc:
```suggestion
		// Unimplemented.
```

src/handlers/ReactNative.ts
```diff
@@ -671,6 +671,20 @@ export class ReactNative extends HandlerInterface
 		await this._pc.setLocalDescription(answer);
 	}
 
+	async pauseReceiving(
+		// eslint-disable-next-line @typescript-eslint/no-unused-vars
+		localId: string): Promise<void>
+	{
+		// Unimplemented
```
ibc:
```suggestion
		// Unimplemented.
```

src/handlers/ReactNative.ts
```diff
@@ -671,6 +671,20 @@ export class ReactNative extends HandlerInterface
...
+	}
+
+	async resumeReceiving(
+		// eslint-disable-next-line @typescript-eslint/no-unused-vars
+		localId: string): Promise<void>
+	{
+		// Unimplemented
```
ibc:
```suggestion
		// Unimplemented.
```

src/handlers/Safari11.ts
```diff
@@ -739,6 +739,20 @@ export class Safari11 extends HandlerInterface
 		return rtpReceiver.getStats();
 	}
 
+	async pauseReceiving(
+		// eslint-disable-next-line @typescript-eslint/no-unused-vars
+		localId: string): Promise<void>
+	{
+		// Unimplemented
```
ibc:
```suggestion
		// Unimplemented.
```

src/handlers/Safari11.ts
```diff
@@ -739,6 +739,20 @@ export class Safari11 extends HandlerInterface
...
+	}
+
+	async resumeReceiving(
+		// eslint-disable-next-line @typescript-eslint/no-unused-vars
+		localId: string): Promise<void>
+	{
+		// Unimplemented
```
ibc:
```suggestion
		// Unimplemented.
```

ibc (commented):
Just a question and some super cosmetic changes requested. Otherwise looks great.

src/handlers/Chrome74.ts
ggarber: In this case the remoteSdp should be `a=sendonly` (because it is a consumer) and the localSdp (generated with createAnswer) will override it with `a=inactive` 

I re-tested it and this is the screenshot of the renegotiation in webrtc-internals:
Screenshot 2021-09-22 at 12 59 47

ggarber:
Thank you Iñaki for the quick review.  Fixed and replied the question.

Also fyi we tested it yesterday in a big audio-only room with most of the people muted (common scenario) and the savings are ~50% of the cpu usage with this change.

FWIW I think webrtc should handle this internally and don't do any audio processing after 5 seconds without audio packets received, but that's a separate conversation that I'll try to move forward in parallel.

Thx.

ibc:
Great. I'll merge after lunch. Just a suggestion for your service: why do you keep receiving videos that receiver is not gonna render anyway? You can also save bandwidth by telling mediasoup to pause() those consumers in the meantime (we do it at Around).

ggarber:
> Great. I'll merge after lunch. Just a suggestion for your service: why do you keep receiving videos that receiver is not gonna render anyway? You can also save bandwidth by telling mediasoup to pause() those consumers in the meantime (we do it at Around).

We also do that for video.

But this optimization is mostly for audio.     Even if you pause the consumers for audio from the server, webrtc will still keep trying to generate some kind of confort noise locally for those consumers because there is no way for webrtc to know that you have stopped sending packets from the server (it assumes that there is packet loss).

ibc:
mediasoup-client 3.6.42 released. Thanks!

PR #181, opened 2021-10-05
Cleanup mapMidTransceiver when stop sending or receiving

vpalmisano:
When we call `stopSending` and `stopReceiving` we close the corresponding media section, stopping the related `RTCRtpTransceiver` object. We need to remove that object from the `_mapMidTransceiver` Map, avoiding to accumulate stopped transceiver objects.


ibc (approved):
Good catch

ibc:
Published in 3.6.43

PR #184, opened 2021-11-17
Add support Microsoft Edge version 88 or above

mmasaki:
This change adds support for Microsoft Edge version 88 or above.


ibc:
Thanks

PR #185, opened 2021-11-22
Add support for Firefox on iOS 14.3 or above

mmasaki:
This change adds support for Firefox on iOS 14.3 or above.

### DIFF
--- a/src/Device.ts
+++ b/src/Device.ts
@@ -102,6 +102,11 @@ export function detectDevice(): BuiltinHandlerName | undefined
 		{
 			return 'Firefox60';
 		}
+		// Firefox on iOS.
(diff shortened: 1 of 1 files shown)


src/Device.ts
```diff
@@ -102,6 +102,11 @@ export function detectDevice(): BuiltinHandlerName | undefined
 		{
 			return 'Firefox60';
 		}
+		// Firefox on iOS.
+		else if (browser.satisfies({ ios: { OS: '>=14.3', firefox: '>=30.0' }}))
+		{
+			return 'Safari12';
+		}
```
ibc, line 109: Can you please enter in https://bowser-js.github.io/bowser-online/ with that browser and paste here the result?

ibc: I've done it. Merging. Thanks.

mmasaki: Thanks!

PR #191, opened 2022-01-07
Concurrent Consumer creation

jmillan:
Create Consumers concurrently.

When there is an ongoing Consumer creation, store the subsequent Consumer creation tasks and run them at once after the current one is created.

This allows operating RTCPeerConnection a single time for multiple track reception rather than doing it sequentially one by one, with the corresponding speed gain.


msach22:
🔥 🔥

src/Transport.ts
```diff
@@ -19,6 +19,24 @@ interface InternalTransportOptions extends TransportOptions
 	canProduceByKind: CanProduceByKind;
 }
 
+class ConsumerCreationTask
+{
+	consumerOptions: ConsumerOptions;
+	promise: Promise<Consumer>;
+	resolve?: (consumer: Consumer) => void;
+	reject?: (error: Error) => void;
```
ibc, line 27: Shouldn't some of those be `private`?

src/Transport.ts
```diff
@@ -788,6 +775,124 @@ export class Transport extends EnhancedEventEmitter
...
+					const results = await this._handler.receive(optionsList);
+
+					for (let idx=0; idx < results.length; idx++)
+					{
+						const task = pendingConsumerTasks[idx];
+						const result = results[idx];
+						const { id, producerId, kind, rtpParameters, appData } = task.consumerOptions;
+
+						const { localId, rtpReceiver, track } = result;
+						const consumer = new Consumer(
```
ibc:
```suggestion
						const task = pendingConsumerTasks[idx];
						const result = results[idx];
						const { id, producerId, kind, rtpParameters, appData } = task.consumerOptions;
						const { localId, rtpReceiver, track } = result;
						const consumer = new Consumer(
```

src/Transport.ts
```diff
@@ -788,6 +775,124 @@ export class Transport extends EnhancedEventEmitter
 			'transport.consumeData()');
 	}
 
+	// This method is guaranteed to never throw.
+	async _createPendingConsumers(): Promise<void>
+	{
+		this._awaitQueue.push(
+			async () =>
+			{
+				this._consumerCreationInProgress = true;
```
ibc, line 784: Do we really need this flag? Couldn't we just rely on `this._pendingConsumerTasks.length`?

src/Transport.ts
jmillan: nop, we need to access all of them from `Transport`

src/Transport.ts
jmillan: Imagine we are creating the 1st `Consumer`.  Before the `Consumer` creation `Transport.consume()` get's called again. `pendingConsumerTasks.length` would be 0 and we would, thus, not add this task to the array but call `handler.receive` as before.

ibc (approved):

PR #193, opened 2022-01-18
Handlers: ReactNative, fix concurrent track reception

jmillan:
Fix ReactNative handler wrong `streamId` usage.


ibc (commented):
How is this NOT a problem also in other PlanB handlers such as Chrome56, Safari11 etc?

jmillan:
> How is this NOT a problem also in other PlanB handlers such as Chrome56, Safari11 etc?

Looking at it

jmillan:
Because this is React Native specific, where the `streamId` we create contains the suffix: **streamId += `-hack-${utils.generateRandomNumber()}`;**

Only needed there.

ibc (approved):

ibc:
Release please

PR #194, opened 2022-01-23
Consumer/Producer.kind typed to MediaKind instead of string

tdrz:
Sets the return type of kind() getter of both Consumer and Producer to be of type MediaKind instead of string. 

The [API docs ](https://mediasoup.org/documentation/v3/mediasoup-client/api/#consumer-kind) seem right:

> consumer.kind
> The media kind (“audio” or “video”).
> @type MediaKind, read only

> producer.kind
> The media kind (“audio” or “video”).
> @type MediaKind, read only


nazar-pc (approved):
Thanks!

jmillan (approved):
:+1:

ibc (approved):
Thanks

PR #195, opened 2022-01-24
bug fix：some exception on handle.receive().

brainwith:
without mid in rtpParameters (this always happend on older mediasoup). When we use multiple audio or video consumers on a single receive transport(UnifiedPlan mode), If we call internalHandler.Receive() and internalHandler.StopReceiving() in multiple times, then internalHandler.receive() will throw an exception. The reason is that multiple calls will lead to disorder of mid, which makes RemoteSdp.receive() incorrectly judge to use PlanB in , resulting in receive exception

PR #196, opened 2022-01-25
Handlers: fix 'receive() when no mid is provided'

jmillan:
Fixes #195


jmillan:
@ibc, similar problem to what happened with react native. We need to store the `localId` for later usage.

ibc (approved):

ibc:
Nice. Please make a release when merged.

PR #198, opened 2022-02-08
fix broswer mid exception

brainwith:
merge bug fix


src/handlers/Chrome70.ts
```diff
@@ -677,7 +677,8 @@ export class Chrome70 extends HandlerInterface
 
 			logger.debug('receive() [trackId:%s, kind:%s]', trackId, kind);
 
-			const localId = rtpParameters.mid || String(this._mapMidTransceiver.size);
+			const nextMSid = this._remoteSdp!.getNextMediaSectionIdx();
+			const localId = rtpParameters.mid || nextMSid.reuseMid || String(nextMSid.idx);
```
ibc, line 681: I fail to understand why we are using (or maybe using) reuseMid here.

test/test.js
```diff
@@ -20,6 +20,9 @@ const utils = require('../lib/utils');
 const { RemoteSdp } = require('../lib/handlers/sdp/RemoteSdp');
 const { FakeHandler } = require('../lib/handlers/FakeHandler');
 const fakeParameters = require('./fakeParameters');
+const uuidv4 = require('uuid/v4');
```
ibc: External deps must be on top.

test/test.js
```diff
@@ -20,6 +20,9 @@ const utils = require('../lib/utils');
...
+
```
ibc: Please no blank line here.

test/test.js
```diff
@@ -875,6 +878,180 @@ test('transport.consume() with a non object appData rejects with TypeError', asy
 		.toThrow(TypeError);
 }, 500);
 
+test('transport.consume() with mutiple video and single audio(fix mid exception)', () =>
```
ibc, line 880: Space required before (

jmillan:
Without any kind of description this PR will be closed.

ibc (changes requested):
Some cosmetic changes requested and a question. I'd like that @jmillan jumps into this since I don't really understand the bug being fixed here.

test/test.js
brainwith: I don't understand, this doesn't seem to have spaces at all

src/handlers/Chrome70.ts
brainwith: in #195 , because getNextMediaSectionIdx() looks like it will return a closed mediaSection, so reuse it

test/test.js
brainwith: checked!

test/test.js
brainwith: checked!

brainwith:
> Without any kind of description this PR will be closed.

hi @jmillan  this PR fix #195

brainwith (commented):
cosmetic

test/test.js
```diff
@@ -875,6 +877,180 @@ test('transport.consume() with a non object appData rejects with TypeError', asy
...
+{
```
jmillan, line 881: It would be great having some meaningful `expect`  in this test.

jmillan: I know that not throwing is an assert in this case, but adding few `expect`s would be more future proof.

brainwith: I have added some exception tests

brainwith:
close to merge

brainwith:
reopen

PR #201, opened 2022-02-25
add WeChat inline browser support on iOS whith version >=14.3

alucardpj:
WeChat is a very popular IM App in China, so there is a lot of WebRTC scene on WeChat inline browser. The iOS with version >=14.3 has already supported WebRTC in WkWebView, this patch add the mediasoup-client support in iOS WeChat >= 14.3.

### DIFF
--- a/src/Device.ts
+++ b/src/Device.ts
@@ -107,6 +107,10 @@ export function detectDevice(): BuiltinHandlerName | undefined
 		{
 			return 'Safari12';
 		}
+		// WeChat inline browser on iOS.
+		else if (browser.satisfies({ ios: { OS: '>=14.3', wechat: '>=0' } })) {
+		    return 'Safari12';
+		}
 		// Safari with Unified-Plan support enabled.
 		else if (
 			browser.satisfies({ safari: '>=12.0' }) &&


src/Device.ts
```diff
@@ -107,6 +107,10 @@ export function detectDevice(): BuiltinHandlerName | undefined
 		{
 			return 'Safari12';
 		}
+		// WeChat inline browser on iOS.
+		else if (browser.satisfies({ ios: { OS: '>=14.3', wechat: '>=0' } })) {
```
ibc, line 111: Why not just ios OS: iOS WeChat >= 14.3? Why should we need to deal with every app safari based browser?

alucardpj: As I said, WeChat is the most popular IM App in China, nearly everyone uses this app for daily communication, it is one of the major browsers in fact just like safari/chrome(maybe more important than them). So in my opinion this individual rule for WeChat is a significant improvement. 
As for the rule itself, because WeChat just use WkWebView for its inline browser, and only iOS >=14.3 the WkWebView supports getusermedia API as well as other WebRTC APIs. In this case, the WeChat's version is not the main constraint, so I think this rule is just OK.

alucardpj: Maybe I made a confusion statement, I mean iOS >=14.3, not WeChat >=14.3, its my fault :)

ibc:
I am closing this since we no longer use bowser library but ua-parser.

ibc:
I hope latest changes in browser detection resolved this issue. Latest PR is this one not merged yet:

https://github.com/versatica/mediasoup-client/pull/339/commits

PR #203, opened 2022-03-16
Handlers: add the ability to pause/resume Consumers concurrently

jmillan:
For the same reason we added concurrency for `handler.receive()`. Adding concurrency for pause/resume avoids multiple SDP renegotiation done one by one which can take a lot of time when there are many consumers.


msach22:
Nice work 😄

ibc (approved):

ibc:
Release?

jmillan:
> Release?

Yes, done

zhangshaoju1987:
This is so great ! 
Does react-native handler also support consumer concurrent?

PR #200, opened 2022-02-23
Improve typings for event emitters

unao:
As a logical continuation of https://github.com/versatica/mediasoup/pull/764 would like to add analogous types for mediasoup client.

I introduced some _implicit_ assumption around `safeEmitAsPromise`: if corresponding type is not specified it allows to pass any arguments otherwise it must be a single argument conforming to the first declared type.

There is still some room left for improvements around `AppData` (for preventing unsafe cast) - if you were interested I could add it once this is merged in.

PR #208, opened 2022-05-19
Better syntax for appData

ibc:


jmillan (approved):
:+1:

unao:
Changing `any` to `Record<string, unknown>` is a breaking change. 

Not sure what exactly is the benefit of forcing  `Record<string, unknown>` of `appData` (and not simply `unknown`).  My understanding was `appData` is totally opaque for the mediasoup-client lib.

This could be improved substantially by specifying a generic type `A extends Record<string, unknown> = Record<string, unknown>` or imho even better `A = unknown`. Without having generic `appData` it is non-trivial to integrate this lib into a typescript codebase.

Actually I wanted to suggest these changes as I found them quite needed, but waiting for 3 months for this PR to be reviewed: https://github.com/versatica/mediasoup-client/pull/200

ibc:
appData is defined as an object that ALWAYS exists and the app can just set string values with any value.

unao:
I think more people see beneficial to have `appData` properly typed with generics: https://github.com/versatica/mediasoup/issues/832.

PR #210, opened 2022-07-15
Fix Consumer creation/pause/resume in batch

douglaseel:
Currently `Transport` is using an AwaitQueue to do tasks related to peerConnection stuff. When there is a task in progress (like a consumer creation or ice stuff) and we try to pause/resume other consumers, it will add a new task to the queue for each `consumer.pause()/consumer.resume()` call instead a single one for all them. 

Why?

The `this._consumerResumeInProgress` and `this._consumerPauseInProgress` are only set as true when we start to run that task. (See https://github.com/versatica/mediasoup-client/blob/v3/src/Transport.ts#L981 and https://github.com/versatica/mediasoup-client/blob/v3/src/Transport.ts#L943)

It means that if we didn't start to process any pause/resume yet, it will add new tasks. For example:

- If we call `transport.consume()` and before this promise resolves we call `consumer1.pause()`, `consumer2.pause()`, `consumer3.pause()` it will push three new different tasks in the queue because `this._consumerPauseInProgress` is still false.

- In the first task related to pause consumers, it will call setRemoteDescription and update the state for all tracks. The other two tasks will call setRemoteDescription without changing  anything. And this could consume a lot of time and could delay another relevant tasks in the queue.

### DIFF
--- a/src/Transport.ts
+++ b/src/Transport.ts
@@ -937,11 +937,11 @@ export class Transport extends EnhancedEventEmitter<TransportEvents>
 
 	_pausePendingConsumers()
 	{
+		this._consumerPauseInProgress = true;
+
 		this._awaitQueue.push(
 			async () =>
 			{
-				this._consumerPauseInProgress = true;
-
 				const pendingPauseConsumers = Array.from(this._pendingPauseConsumers.values());
 
 				// Clear pending pause Consumer map.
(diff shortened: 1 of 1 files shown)


jmillan:
Looks good @douglaseel 👍 , I'll further comment       in the following days.

jmillan:
Same applies to Consumer creation. Please add this too:

```diff
diff --git a/src/Transport.ts b/src/Transport.ts
index f10f46c..a4ff13e 100644
--- a/src/Transport.ts
+++ b/src/Transport.ts
@@ -786,11 +786,11 @@ export class Transport extends EnhancedEventEmitter
        // This method is guaranteed to never throw.
        async _createPendingConsumers(): Promise<void>
        {
+               this._consumerCreationInProgress = true;
+
                this._awaitQueue.push(
                        async () =>
                        {
-                               this._consumerCreationInProgress = true;
-
                                const pendingConsumerTasks = [ ...this._pendingConsumerTasks ];

                                // Clear pending Consumer tasks.
```

jmillan (approved):
:+1: Nice

ibc (approved):
Definitely good catch. Will merge soon. Thanks!

ibc:
Released in mediasoup-client 3.6.54

PR #212, opened 2022-07-21
Avoiding unnecessary peer connection renegotiation in consumer resume/pause methods

douglaseel:
This PR add some checks to avoid unnecessary peer connection renegotiation when:

1 - `consumer.pause()` is called for an already `paused` consumer;
2 - `consumer.resume()` is called for an already `unpaused` consumer;
3 - `_pendingPauseConsumers` Map in Transport is empty and the task to pause consumers in batch is called. It could happen when we call `consumer.pause()` and `consumer.resume()` after and the first task has not run yet;
4 - `_pendingResumeConsumers` Map in Transport is empty and the task to resume consumers in batch is called.  It could happen when we call `consumer.resume()` and `consumer.pause()` after and the first task has not run yet;


src/Consumer.ts
```diff
@@ -283,6 +290,13 @@ export class Consumer extends EnhancedEventEmitter<ConsumerEvents>
 			return;
 		}
 
+		if (!this._paused)
+		{
+			logger.debug('pause() | Consumer is already resumed');
```
ibc: "resume()"

src/Transport.ts
```diff
@@ -848,6 +848,13 @@ export class Transport extends EnhancedEventEmitter<TransportEvents>
 					});
 				}
 
+				if (optionsList.length === 0)
+				{
+					logger.debug('_createPendingConsumers() | There is no consumer to be created');
```
ibc: "there"

src/Transport.ts
```diff
@@ -947,22 +954,30 @@ export class Transport extends EnhancedEventEmitter<TransportEvents>
 				// Clear pending pause Consumer map.
 				this._pendingPauseConsumers.clear();
 
+				const localIds = pendingPauseConsumers
+					.map((consumer) => consumer.localId);
+
+				if (localIds.length === 0)
+				{
+					logger.debug('_pausePendingConsumers() | There is no consumer to be paused');
```
ibc: "there"

src/Transport.ts
```diff
@@ -985,6 +1000,16 @@ export class Transport extends EnhancedEventEmitter<TransportEvents>
 				// Clear pending resume Consumer map.
 				this._pendingResumeConsumers.clear();
 
+				const localIds = pendingResumeConsumers
+					.map((consumer) => consumer.localId);
+
+				if (localIds.length === 0)
+				{
+					logger.debug('_resumePendingConsumers() | There is no consumer to be resumed');
```
ibc: "there"

ibc (commented):
Cosmetic changes requested.

Looks good but I'll let @jmillan merge it because I'll go to the beach.

src/Transport.ts
```diff
@@ -848,6 +848,13 @@ export class Transport extends EnhancedEventEmitter<TransportEvents>
 					});
 				}
 
+				if (optionsList.length === 0)
+				{
+					logger.debug('_createPendingConsumers() | there is no consumer to be created');
```
ibc: "Consumer"

src/Transport.ts
```diff
@@ -947,22 +954,30 @@ export class Transport extends EnhancedEventEmitter<TransportEvents>
 				// Clear pending pause Consumer map.
 				this._pendingPauseConsumers.clear();
 
+				const localIds = pendingPauseConsumers
+					.map((consumer) => consumer.localId);
+
+				if (localIds.length === 0)
+				{
+					logger.debug('_pausePendingConsumers() | there is no consumer to be paused');
```
ibc: "Consumer"

src/Transport.ts
```diff
@@ -985,6 +1000,16 @@ export class Transport extends EnhancedEventEmitter<TransportEvents>
 				// Clear pending resume Consumer map.
 				this._pendingResumeConsumers.clear();
 
+				const localIds = pendingResumeConsumers
+					.map((consumer) => consumer.localId);
+
+				if (localIds.length === 0)
+				{
+					logger.debug('_resumePendingConsumers() | there is no consumer to be resumed');
```
ibc: "Consumer"

src/Transport.ts
jmillan: This can be moved to the beginning for the method to avoid superfluous `map()` exection, for clarity:

 ```ts
if (this._pendingPauseConsumers.size === 0)
{
  logger.debug('_pausePendingConsumers() | there is no consumer to be paused');

  return;
}
```

Same with the rest of places where this is done.

ibc:
Reminder that this PR may also have to do the same for new consumer produce stuff added today by Jose.

douglaseel:
> Reminder that this PR may also have to do the same for new consumer produce stuff added today by Jose.

Done!

jmillan:
Super 🎉

PR #214, opened 2022-07-24
Fix typescript build error

mmasaki:
Fix typescript build error as below.

```
src/DataConsumer.ts:141:3 - error TS2322: Type 'string' is not assignable to type 'BinaryType'.

141     return this._dataChannel.binaryType;
```


nazar-pc:
I don't think you're supposed to use `src` anyway, use `lib` instead.

ibc:
Sorry for the super delay. However in latest versions of TS this is not an issue anymore.

PR #215, opened 2022-07-27
Effectively disable encoder on Producer pause

jmillan:
..., by setting the media description direction to 'inactive'  on pause.

Changes:
* `zeroRtpOnPause` does not replace the track with `null` but properly negotiates with `a=inactive`.
*  new `pauseSending()` `resumeSending()` methods have been created in handlers, and implemented in those where `pauseReceiving()` and `resumeReceiving()` is implemented, for consistency.
* This PR does not contain the ability to concurrently pause/resume Producers, on purpose. (I want to refactor the tests before doing that)

@ibc, with this change I don't know if `zeroRtpOnPause` has sense anymore since by default IMHO we should be pausing/resuming Producers by this new way, so IMHO we can deprecate it in an upcoming version in the future and always pause/resume Producers this way.


ibc (approved):
lovely. Remember we have to enable the `zeroRtpOnPause` flag at work XD

tdrz:
Without meaning to hijack this PR: is this also implemented in libmediasoupclient?

ibc:
> Without meaning to hijack this PR: is this also implemented in libmediasoupclient?

I don't think so

PR #216, opened 2022-08-24
Fix replacetrack when zeroRtpOnPause is enabled

ggarber:
When _zeroRtpOnPause is enabled replacing the track while the producer is paused doesn't work because the RtpSender track is not replaced.    Is there a reason for that behaviour different than when _zeroRtpOnPause is disabled?

When unmuting a video participant we are unpausing the producer and replacing the track with a new one from getUserMedia and we do those operations in no specific order so it can happen that the producer is still paused when replacing the track.

This PR makes the replacetrack call happen always no matter what while before it was happening always apart from that specific case.   Maybe there is a better solution or some side effects of this change so it would be good to know what you think.

### DIFF
--- a/src/Producer.ts
+++ b/src/Producer.ts
@@ -419,18 +419,15 @@ export class Producer extends EnhancedEventEmitter<ProducerEvents>
 			return;
 		}
 
-		if (!this._zeroRtpOnPause || !this._paused)
+		await new Promise<void>((resolve, reject) =>
 		{
-			await new Promise<void>((resolve, reject) =>
-			{
-				this.safeEmit(
-					'@replacetrack',
-					track,
-					resolve,
-					reject
-				);
-			});
-		}
+			this.safeEmit(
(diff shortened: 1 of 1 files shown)


src/Producer.ts
```diff
@@ -419,18 +419,15 @@ export class Producer extends EnhancedEventEmitter<ProducerEvents>
 			return;
 		}
 
-		if (!this._zeroRtpOnPause || !this._paused)
+		await new Promise<void>((resolve, reject) =>
```
ibc, line 422: Súper busy until next week but I fail to see why the if condition has beer entirely removed. We have to check if zeroRtpOnPause flag is set if I am not wrong. Also it looks like there are lint problems but I may be wrong. npm run lint.

ibc: it's ok

ibc:
Released in 3.6.57

PR #217, opened 2022-09-28
Add support for sending abs-capture-time RTP extension

ibc:
Since recent versions, Chromium/Chrome has support for sending the `abs-capture-time` RTP extension, however it lacks some stuff (read the issue in the link below):

> - Negotiate the header extension.
> - Collect capture time for audio and video and have the info sent with the header extension.
> - Receive the header extension and use its info.

_NOTE:_ If the second bullet is true it means that, even forcing the extension to be sent via SDP munging, it's not sending useful content ¯\_(ツ)_/¯

This PR  forces negotiation of the `abs-capture-time` for sending by means of local SDP offer munging.

- https://bugs.chromium.org/p/webrtc/issues/detail?id=10739
- https://webrtc.googlesource.com/src/+/refs/heads/main/docs/native-code/rtp-hdrext/abs-capture-time


jmillan:
Does it need be exposed in the public API?

Why not simply always force it in `chrome74.ts` until it's done by the implementation itself?

ibc:
> Does it need be exposed in the public API?
> 
> Why not simply always force it in `chrome74.ts` until it's done by the implementation itself?

No need to send an extra header (it means more bandwidth usage) if it's not gonna be consumed.

jmillan:
> No need to send an extra header (it means more bandwidth usage) if it's not gonna be consumed.

But by the time Chrome decides to expose it always (as it does with the rest of them), it will be sent always as the rest of unsolicited extra headers, am I wrong?

ibc:
> > No need to send an extra header (it means more bandwidth usage) if it's not gonna be consumed.
> 
> But by the time Chrome decides to expose it always (as it does with the rest of them), it will be sent always as the rest of unsolicited extra headers, am I wrong?

Right. I'll remove the param.

jmillan (approved):

ibc:
finally nobody seems to need this feature so I'll convert to a draft.

fippo:
> This PR forces negotiation of the abs-capture-time for sending by means of local SDP offer munging.

Your SDP O/A foo is weak son :trollface: 
It is entirely sufficient to add that extmap to the remote answer to cause the sender to send the extension as demonstrated by [this fiddle](https://jsfiddle.net/fippo/ymuaqv85/5/). O/A with quirks!

ibc:
> It is entirely sufficient to add that extmap to the remote answer to cause the sender to send the extension

It shouldn't work ¯\_(ツ)_/¯

ibc:
This is done in another PR.

PR #219, opened 2022-10-19
Fix answer media direction when pausing sending

ggarber:
When pausing sending a media track the Offer generated by the browser includes "a=inactive" but the Answer generated by mediasoup-client still has "a=recvonly" that is not legal.

This is apparently not a problem with Chrome but with Firefox in some cases we see the Answer being rejected and the producers failing.

This PR includes the fix to Firefox for now but will have the same fix for the other handlers too if the solution proposed makes sense.


lib/handlers/sdp/RemoteSdp.js
```diff
@@ -267,6 +271,13 @@ class RemoteSdp {
             this._sdpObject.media[idx] = newMediaSection.getObject();
         }
     }
+    _findMediaSection(mid) {
```
ibc, line 274: I don't understand why we may need this method, we are supposed to already be able to identify the media section somehow else in many other methods. Please remember that we are ready for Unified and Plan B in RemoteSdp class.

ggarber: I created this private helper method only to avoid the code duplication.   This code to find a mediaSection from a mid was duplicated in `closeMediaSection()` and `disableMediaSection()`.

ggarber:
@ibc Apart from the style/code comments, does the issue and fix make sense?    Just trying to confirm it before updating all the handlers :)

ibc:
Please let me properly review it tomorrow (reminder created). Too busy today.

ibc (commented):
Ok, I like it. So will you update all handlers before meeting this? I'm super fine with that.

NOTE: don't do it in plan B based handlers.

ggarber:
Thank you! Rest of handlers requiring this change updated and I did some more testing with Chrome to see that nothing is apparently broken with this change.

ibc (approved):
Nice. Will release a new version this weekend.

ibc:
NOTE: There are linting errors in this PR. I'm fixing them in v3 branch. Please always run `npm run lint` and `npm run test`.
CleanShot-2022-10-22-at-19 51 18@2x

ibc:
Version 3.6.58 released.

PR #227, opened 2022-11-29
Migrate ReactNative pc.addStream to pc.addTrack

charlyBerthet:
RTCPeerConnection.addStream is deprecated (https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/addStream).  Let's migrate to the new addTrack API.

Because it is deprecated, right now pc.addStream is undefined in ReactNative so we cannot use mediasoupSendTransport.produce({ track }).

### DIFF
--- a/src/handlers/ReactNative.ts
+++ b/src/handlers/ReactNative.ts
@@ -314,7 +314,7 @@ export class ReactNative extends HandlerInterface
 		}
 
 		this._sendStream.addTrack(track);
-		this._pc.addStream(this._sendStream);
+		this._pc.addTrack(track, this._sendStream);
 
 		let offer = await this._pc.createOffer();
 		let localSdpObject = sdpTransform.parse(offer.sdp);


ibc:
Thanks, but actually we have to migrate to Unified Plan anyway (see the issue in GH) so must do much more than just this.

BTW are you in the mediasoup support forum? I'll make an announcement post there telling that I'll upgrade ReactNative handler to Unified Plan but need someone testing it (right now I cannot test it locally). Would you mind testing it when done?

ibc:
Announcement here: https://mediasoup.discourse.group/t/need-testers-for-reactnative-unified-plan/4739

Closing this PR in favor of the incoming one with Unified Plan full support.

ibc:
PR with Unified Plan support done and needs testing: https://github.com/versatica/mediasoup-client/pull/228

charlyBerthet:
@ibc thanks! Sure I can test it. It's the first time I am using mediasoup so I'll try to do my best 👊 I saw your PR https://github.com/versatica/mediasoup-client/pull/228

PR #241, opened 2023-01-27
Update scalability mode for latest Chrome and Firefox

jmillan:
Temporal layers per stream were downgraded from three to two time ago.

* Safari Tech Preview still uses 3 temporal layers.
* I did not test ReactNative, so I'm keeping it as it was.

https://bugs.chromium.org/p/webrtc/issues/detail?id=11366
https://groups.google.com/g/discuss-webrtc/c/N1sMEBJhOz4

### DIFF
--- a/src/handlers/Chrome74.ts
+++ b/src/handlers/Chrome74.ts
@@ -421,7 +421,7 @@ export class Chrome74 extends HandlerInterface
 		{
 			for (const encoding of sendingRtpParameters.encodings)
 			{
-				encoding.scalabilityMode = 'S1T3';
+				encoding.scalabilityMode = 'S1T2';
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


ibc (approved):
The ScalabilityMode API is about to be enabled in libwebrtc (a Google dev told me):

- https://www.w3.org/TR/webrtc-svc/
- https://github.com/w3c/webrtc-svc/blob/main/explainer.md

So I hope we can get rid of those hardcoded values.

vpalmisano:
FYI Chrome on Linux still uses 3 temporal layers by default.

ibc:
> FYI Chrome on Linux still uses 3 temporal layers by default.

Definitely we don't want to hardcode OS dependant guessed values, specially now that libwebrtc guys are working on implementing the `scalabilityMode` API in WebRTC.

However, I've done this: https://github.com/versatica/mediasoup-client/pull/244

PR #242, opened 2023-01-30
Add support for Opus NACK

t-mullen:
Adds support for NACK packets and retransmissions when using Opus. 
This makes a significant difference when packet loss is high. Here's some samples with 40% packet loss.
Before: https://voca.ro/1drKi1WrGBbL
After: https://voca.ro/1c5NmOL4D42u

Both Safari and Chrome support sending and receiving. No issue was observed when enabling this feature on other browsers, but I left it out of the handlers anyways. Edge may be supported but I have not tested it thoughly (may add it in later PR).

Changes summary:
- Added an `opusNack` option to the `ProducerCodecOptions`. This is `false` by default.
- Chrome74 and Safari12 handlers can now inject the `rtcp-fb nack` SDP line via rtcpFeedback.

Depends on server PR: https://github.com/versatica/mediasoup/pull/997


t-mullen:
I see there's conflicts with the built types / JS. Should I not commit these files?

lib/Producer.d.ts
```diff
@@ -14,6 +14,7 @@ export declare type ProducerCodecOptions = {
     opusStereo?: boolean;
     opusFec?: boolean;
     opusDtx?: boolean;
+    opusNack?: boolean;
```
ggarber: Rest of opusXXXX parameters are really opus parameters but the nack one is not a codec parameter but a transport one.   It shouldn't be a codec option  IMO.

ibc: The thing here is that, in order to make libwebrtc offer NACK in OPUS, we need to mangle the SDP offer.

Do you mean that it should be codec agnostic, something such as `forceNack` instead?

ibc:
Please remove `lib` folder in your local fork. `lib` folder was removed from mediasoup-client repo some months ago.

src/handlers/sdp/MediaSection.ts
```diff
@@ -248,6 +249,10 @@ export class AnswerMediaSection extends MediaSection
 									codecParameters.usedtx = opusDtx ? 1 : 0;
 								}
 
+								if (!opusNack) {
+									offerCodec!.rtcpFeedback = offerCodec?.rtcpFeedback?.filter(fb => fb.type !== 'nack')
+								}
```
ibc, line 254: AFAIU it should also modify `codecParameters` since those are the ones that will be signaled to mediasoup server.
But I don't really understand what this code is doing. I read:

"If `opusNack` is not set, then remove "nack" from its supported RTCP feedbacks"

Why? Here we are supposed to add features rather than removing them.

lib/Producer.d.ts
ggarber: Yes, I think should be codec agnostic (because it is codec agnostic).   I think this setting should also allow disabling nacks for video if needed even if it is not a practical use case.   `forceNack` or `enableNack`.

ggarber: Also if there is a different place for it instead of ProducerCodecOption structure it would be better given that this is not really a codec parameter.   But I'm not sure if this is doable in the current design.

ibc: RTCP feedbacks are somehow "codec options" so I think the option is fine here.

However I need to properly review this PR once I get time since I'm not fully convinced about how it's done. So please be patient.

src/handlers/sdp/MediaSection.ts
t-mullen: The `{ type: 'nack' }` is added into the native RTP capabilities for Chrome74 and Safari12, so we need to remove them if the application doesn't want Opus NACK.

I could invert this code to add NACK here instead and not add NACK into the native RTP capabilities, but then it would need to be added in a few other places that the native RTP capabilities are used. This was the most minimal change I could think of.

lib/Producer.d.ts
fippo: Why do you folks think that bad SDP munging is required? On the [sample page](https://webrtc.github.io/samples/src/content/peerconnection/audio/) try this (with loss)
```
await pc1.setLocalDescription();
await pc2.setRemoteDescription({type: 'offer', sdp: pc1.localDescription.sdp +  'a=rtcp-fb:111 nack\r\n'});
await pc2.setLocalDescription();
await pc1.setRemoteDescription({type: 'answer', sdp: pc2.localDescription.sdp +  'a=rtcp-fb:111 nack\r\n'});
```
and then observe the `retransmittedPacketsSent` on the outbound-rtp stats increase.
No munging, just some SDP manipulation which is totally allowed.

ibc: What? We do it much better by modifying the specific m= section we want to affect and not just the last one or the whole SDP.

fippo: The point is that you don't even need clientside support. The only browser not supporting this is Firefox and what happens if the server announces nack is simply that Firefox won't do resends. Easier to infer from the UA than putting it into caps

ibc: Definitely we don't want that mediasoup sends NACK to a device that doesn't support NACK for the given codec.

ibc:
I'm checking this and I'm afraid I cannot accept this PR as it is, there a too many missing things such as the ability to not just send OPUS retransmissions from client to mediasoup but also from mediasoup to client (this requires changes in mediasoup and many more changes in mediasoup-client). Honestly this is not just avoid adding those missing things. This PR doesn't follow the mediasoup style of doing things and I'll implement this feature from scratch when possible. I'll let this PR open as reference.

ibc:
I have a question here: Should we enable RTX for Opus as well? Does somebody know if libwebrtc would honor it? It would mean that we should munge the SDP offer and add a RTX codec for Opus (and also in receiver side because this feature cannot be just about OPUS retransmissions from client to server).

fippo:
libwebrtc does not implement rtx for audio (which has implications on the jitter buffer delay)

ibc:
Similar PR here https://github.com/versatica/mediasoup-client/pull/249

PR #243, opened 2023-02-02
fix: Improve batching behaviour with parallel calls to consume()

t-mullen:
When creating multiple consumers in parallel like this:
```
await Promise.all([
  transport.consume(/* ... */), 
  transport.consume(/* ... */), 
  // ...
])
```
The awaitQueue will be pushed a consumer creation task with the first consumer, then another with all the other consumers. This inefficient batching slows down consumer creation.

To improve this, this PR waits for any existing microtasks to finish before creating the first task. This results in at most one task pushed to the awaitQueue for any calls to consume() still in the microtask queue. The same optimization was applied to pausing/resuming consumers.

### DIFF
--- a/src/Transport.ts
+++ b/src/Transport.ts
@@ -674,6 +674,10 @@ export class Transport extends EnhancedEventEmitter<TransportEvents>
 		// There is no Consumer creation in progress, create it now.
 		queueMicrotask(() => 
 		{
+			if (this._closed) {
+				throw new InvalidStateError('closed');
+			}
+
 			if (this._consumerCreationInProgress === false)
 			{
 				this.createPendingConsumers();
--- a/src/Transport.ts
+++ b/src/Transport.ts
@@ -11,6 +11,7 @@ import { DataProducer, DataProducerOptions } from './DataProducer';
 import { DataConsumer, DataConsumerOptions } from './DataConsumer';
 import { RtpParameters, MediaKind } from './RtpParameters';
 import { SctpParameters, SctpStreamParameters } from './SctpParameters';
+import queueMicrotask from 'queue-microtask';
 
 const logger = new Logger('Transport');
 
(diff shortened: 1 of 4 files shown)


ibc:
Makes sense. Thanks. Will entirely review it on next week.

jmillan (approved):
Nice :-)

src/Transport.ts
```diff
@@ -1249,10 +1256,13 @@ export class Transport extends EnhancedEventEmitter<TransportEvents>
 			this._pendingResumeConsumers.set(consumer.id, consumer);
 
 			// There is no Consumer resume in progress, do it now.
-			if (this._consumerResumeInProgress === false)
+			queueMicrotask(() => 
 			{
-				this.resumePendingConsumers();
-			}
+				if (this._consumerResumeInProgress === false)
```
ibc, line 1265: Here, at the top of the callback, we must check `this._closed` and `throw new InvalidStateError('closed');` if closed.

src/tests/test.ts
```diff
@@ -787,6 +788,55 @@ test('transport.consume() succeeds', async () =>
...
+				{
+					resolve();
+				}
+			});
+		});
+	};
+	const allConsumersCreated = Promise.all([
+		waitForConsumer(videoConsumerRemoteParameters1.id),
+		waitForConsumer(videoConsumerRemoteParameters2.id)
+	]);
```
ibc: cosmetic

```suggestion

	const allConsumersCreated = Promise.all(
		[
			waitForConsumer(videoConsumerRemoteParameters1.id),
			waitForConsumer(videoConsumerRemoteParameters2.id)
		]);
```

ibc (changes requested):

ibc:
This maybe unrelated to this PR:

We are not checking `this._closed` flag in `Transport.ts` when it comes to process enqueued tasks. Let me do this.

src/Transport.ts
ibc: This should be done in all usages of `queueMicrotask()`, not just in this one.

ibc (changes requested):

ibc:
CI is failing: Please run `npm run lint`

ibc:
I will merge and fix locally.

ibc:
Released in 3.6.76. Thanks.

PR #244, opened 2023-02-06
Honor given encoding.scalabilityMode

ibc:
`transport.produce()`: If the app includes  `scalabilityMode` field in `encodings`, honor given number of temporal layers.

Maybe important thing here:

- Imagine the app passes `scalabilityMode: 'S2T2'.
- We honor it. We are in Safari (which as per today generates **3 temporal layers**).
- In mediasoup Producer stats we'll just see stats for 2 temporal layers (since we signaled `S2T2` to mediasoup `produce()`).
- Are we discarding temporal layer 2 in mediasoup???? NOTE: This is another topic anyway.

### DIFF
--- a/src/handlers/Chrome74.ts
+++ b/src/handlers/Chrome74.ts
@@ -433,7 +433,16 @@ export class Chrome74 extends HandlerInterface
 		{
 			for (const encoding of sendingRtpParameters.encodings)
 			{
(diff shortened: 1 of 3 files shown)


ibc:
CC @vpalmisano

jmillan (approved):

ibc:
Released in 3.6.77

PR #245, opened 2023-02-06
Persist changes made to paused states of spatial layers while using simulcast

thebongy:
Currently, if `setMaxSpatialLayer()` or `setRtpEncodingParameters()` are used to set `active` state for simulcast spatial layers, these changes only persist until the next time some other operation on the same transport calls `setRemoteDescription`. After this, the layers are resumed again even if they were supposed to be paused.

This is reproducible on the mediasoup demo for example, by clicking on "down" to disable a few simulcast layers, and then clicking "Restart Ice", which will trigger a setRemoteDescription() on the transport. Then, the layers which are disabled will get enabled again. "Restart Ice" is just an example, but this is also reproducible on other changes on the transport, for example, making new producers.

This appears to happen because the RemoteSDP doesn't persist the paused and resumed state of each `rid` when we make a setMaxSpatialLayer / setRtpEncodingParameters, and they still consider all layers as unpaused like the following:
```
a=simulcast:recv r0;r1;r2
```

This PR adds a change to correctly modify the remote SDP (and add a `~` against paused layers) when layers are changed for handlers which don't use legacy simulcast, so that the changes are correctly persisted regardless of changes that are made on the transport.


ibc:
This is a good catch, but I don't love the way it's implemented. Let me come back to this on next days. Thanks a lot.

thebongy:
Just wanted to add:

The changes use `any` in a few places because the typings for sdp-transform seem to be wrong, I have opened another seperate PR for that (https://github.com/DefinitelyTyped/DefinitelyTyped/pull/64119)

ibc:
Sure, I didn't mean that. I understand it. Too busy these days so let me please come back to this on next week.

ibc (commented):
To be clear, problems in this PR are:

- SDP utils should be in handlers/sdp/ folder.
- Here we are adding a SDP O/A by calling pc.setLocal/RemoteDescription() within a method in the handler. It can collide with other SDP O/A and fail. That's why all these actions that need SDP O/A must consist on the Producer method emitting an event to Transport.ts and Transport.ts **enqueing** the corresponding call to the handler, as we do with pause() method for example.

More on next week.

thebongy:
> SDP utils should be in handlers/sdp/ folder

This makes sense, I have made a commit to move it to the existing unifiedPlan file.

> Here we are adding a SDP O/A by calling pc.setLocal/RemoteDescription() within a method in the handler. It can collide with other SDP O/A and fail. That's why all these actions that need SDP O/A must consist on the Producer method emitting an event to Transport.ts and Transport.ts enqueing the corresponding call to the handler, as we do with pause() method for example.

Sorry if I didn't understand correctly, but isn't setMaxSpatialLayer and setRtpEncodingParameters already following this flow? 

For example, setMaxSpatialLayer emits an event from producer here (https://github.com/thebongy/mediasoup-client/blob/1fd6448fffe25ca42f1bf6c9416c649a7d0767f3/src/Producer.ts#L467-L475), and transport enqueues the handler method here (https://github.com/thebongy/mediasoup-client/blob/0a9bbd480059ebc85cde75fead46283ddebc2043/src/Transport.ts#L1278-L1286).



Also, please take your time, there is absolutely no hurry in reviewing this. Thank you for having a look at the PR!

ibc:
You are right. Sorry.

ibc:
I pulled this branch but couldn't make changes since it's protected so I ended cloning it in this repository. I've "improved" the code (or made it more mediasoup-client style) and added more handlers. Let's please handle this task in the new PR and tons of thanks for your amazing work.

https://github.com/versatica/mediasoup-client/pull/246

PR #246, opened 2023-02-14
Persist changes in sending encodings

ibc:
Originally done by @thebongy here: https://github.com/versatica/mediasoup-client/pull/245

Cosmetic changes done and more handlers included.

### Steps to verify it

- Open mediasoup-demo.
- Enable mic and webcam (VP8 so with simulcast).
- In local view, enable "controls" and decrease max spatial layer.
- In console run `CC.disableMic()`.
- Check sending webcam spatial layers. It should be the same as before.


ibc:
Merging and releasing.

ibc:
Released in 3.6.78, thanks

jmillan (commented):
:+1:

PR #247, opened 2023-02-14
fix(svc): simulcast/svc broken on M111 because S1T2 is not valid for VP8, use L1T2 instead

thebongy:
Chrome M111 (just became "beta" version yesterday), is rolling out support for SVC, which was previously only behind a flag called "RTCSvcScalabilityMode". 

(Chromium commit: https://chromium.googlesource.com/chromium/src/+/a32ab6f178f6aeb5674afad69c1dfcf83d2a41db%5E%21/#F0)
(I2S:https://groups.google.com/a/chromium.org/g/blink-dev/c/XzygOFuas2A/m/imTtbDqPBgAJ)

This means that chrome will now actually check if the `scalabilityMode` passed in the encodings array is a valid string, because of this patch (https://webrtc-review.googlesource.com/c/src/+/277403). This also means that number of temporal layers are actually respected by chrome.

According to the SVC spec (https://w3c.github.io/webrtc-svc/#scalabilitymodes*):

> For example, VP8 [[RFC6386](https://w3c.github.io/webrtc-svc/#bib-rfc6386)] only supports temporal scalability (e.g. ["L1T2"](https://w3c.github.io/webrtc-svc/#dfn-l1t2), ["L1T3"](https://w3c.github.io/webrtc-svc/#dfn-l1t3))

Therefore, "S" can't be passed for the vp8 codec.

This is also enforced by libwebrtc: 

- for vp8, valid scalability modes are: 
```
ScalabilityMode::kL1T1, ScalabilityMode::kL1T2, ScalabilityMode::kL1T3
```
https://webrtc.googlesource.com/src/+/a1a7c638ece600a0d9e2a1f7d77e1a9050737e90/modules/video_coding/codecs/vp8/vp8_scalability.h
- for vp9/h264, valid scalability modes are documented here (https://webrtc.googlesource.com/src/+/ecfe8da46b22883363a66d0909d7d2246cb13c48/modules/video_coding/svc/create_scalability_structure.cc)

This PR switches from using "S" in scalability modes to "L" to prevent mediasoup from crashing on M111.

**Steps to Reproduce**
1. Open mediasoup demo in chrome dev (Version: M111 or above)
2. Start producing video
3. Console should show folllowing error, and addTransceiver should fail:

image

thanks @CallMeTarush , @palashgo for discovering this

### DIFF
--- a/src/handlers/Chrome74.ts
+++ b/src/handlers/Chrome74.ts
@@ -435,13 +435,13 @@ export class Chrome74 extends HandlerInterface
 			{
 				if (encoding.scalabilityMode)
 				{
-					encoding.scalabilityMode = `S1T${layers.temporalLayers}`;
+					encoding.scalabilityMode = `L1T${layers.temporalLayers}`;
 				}
 				else
 				{
 					// By default Chrome enables 2 temporal layers (not in all OS but
 					// anyway).
-					encoding.scalabilityMode = 'S1T2';
+					encoding.scalabilityMode = 'L1T2';
 				}
 			}
 		}
--- a/src/handlers/Firefox60.ts
+++ b/src/handlers/Firefox60.ts
@@ -435,12 +435,12 @@ export class Firefox60 extends HandlerInterface
 			{
 				if (encoding.scalabilityMode)
 				{
-					encoding.scalabilityMode = `S1T${layers.temporalLayers}`;
+					encoding.scalabilityMode = `L1T${layers.temporalLayers}`;
 				}
 				else
 				{
 					// By default Firefox enables 2 temporal layers.
-					encoding.scalabilityMode = 'S1T2';
+					encoding.scalabilityMode = 'L1T2';
 				}
 			}
 		}


ibc:
Thanks. I was waiting for these changes in chromium. However, some questions before checking this PR:

1. Is it guaranteed that Chrome/Chromium 111 handles `scalabilityMode` in **all** OSs (macOS, Windows, Linux)?

2. This PR replaces "S" with "L" which is ok. However, once we can use `scalabilityMode` there is no reason for those terrible hacks in `Chrome74.ts` (and other handlers) in `send()` method where we mangle the local SDP offer to add more `ssrc` lines and so on. So we need a new `Chrome111.ts` handler that just relies on given `scalabilityMode` for any supported video codec (VP8, VP9 and H264), am I right?

thebongy:
Hi,

> Is it guaranteed that Chrome/Chromium 111 handles scalabilityMode in all OSs (macOS, Windows, Linux)?

From our testing, currently the Chrome74.ts handler file is breaking across all OSs, and this fix does make it work correctly on all three platforms with the L1T2 scalability mode (with the VP8 codec). We have conducted basic tests across multiple devices (and on both chrome beta and stable on each) to verify this.

> This PR replaces "S" with "L" which is ok. However, once we can use scalabilityMode there is no reason for those terrible hacks in Chrome74.ts (and other handlers) in send() method where we mangle the local SDP offer to add more ssrc lines and so on. So we need a new Chrome111.ts handler that just relies on given scalabilityMode for any supported video codec (VP8, VP9 and H264), am I right?

That is correct, we can add a new Chrome111 file. About the legacy simulcast related hacks, I think those can be removed now yes, but this will require extensive testing, and I can make a follow up PR for that.

For the purpose of this PR, I mainly added this because I was confused why does mediasoup have "S1T2" in the Chrome74 in the first place. I tried looking both on SVC spec, chromium source, and libwebrtc source, and find no references to this scalability mode in the code, it seems to have always been "L1T2" instead. That is why I felt that the issue should be fixed in Chrome74 handler as well, unless we had a reason to add "S1T2". Can you give me more context around why was this added, since I feel I may be missing something here? 


PS: Just a note on the urgency of this change, this currently breaks  on chrome beta, which is set to become **chrome stable in 15 days.**. I think if there are no issues on your side, we should have a PSA on the discourse channel to have everyone upgrade their mediasoup client to get the fixes.

ibc:
Merging this and releasing a new version.

I'll take care of `Chrome111` on next weeks.

Thanks a lot.

thebongy:
Thank you!

ibc:
mediasoup-client 3.6.79 release.

ibc:
Do you know which Microsoft Edge version comes or will come with `scalabilityMode` enabled? Basically to fill this:

CleanShot-2023-02-15-at-12 59 52@2x

ibc:
We may need the same for `ReactNativeUnifiedPlan`.

thebongy:
I think edge is basically same as chrome releases according to (https://learn.microsoft.com/en-us/deployedge/microsoft-edge-release-schedule)

thebongy:
Also just a small mistake I made in my previous comment/title of the PR: I believe the crash on mediasoup demo was due to the demo itself passing along S1T2, and not this library technically. The only issue in mediasoup-client itself is that it returns an inconsistent string (S1T2 instead of L1T2) which is a bit misleading, but this is fixed with this merge as well.

I see you have already fixed this in (https://github.com/versatica/mediasoup-demo/commit/d15a859306e1ba5d031cde90d02593e095719cbc), thank you!

thebongy:
@ibc do let me know if there's any way I can help in developing/testing the new chrome 111 handlers you mentioned

vpalmisano:
> Also just a small mistake I made in my previous comment/title of the PR: I believe the crash on mediasoup demo was due to the demo itself passing along S1T2, and not this library technically. The only issue in mediasoup-client itself is that it returns an inconsistent string (S1T2 instead of L1T2) which is a bit misleading, but this is fixed with this merge as well.

Using the current mediasoup demo with Chromium 112.0.5586.0 it works. Did you used some additional query params to set the S1L2 scalability mode?

palashgo:
@vpalmisano I think the fixes were made on the mediasoup demo as well 3-4 hours ago. Did you test before that or after that

vpalmisano:
> @vpalmisano I think the fixes were made on the mediasoup demo as well 3-4 hours ago. Did you test before that or after that

Probably after that change, thanks.

ibc:
https://github.com/versatica/mediasoup-client/pull/248

Tested and works fine.

PR #248, opened 2023-02-15
Chrome111 handler

ibc:
The new `Chrome111` handler uses `scalabilityMode` when in Chrome/Chromium/Edge >= M111 (which supports `encpding.scalabilityMode`), meaning that finally we can decide desired number of spatial and temporal layers.

- Tested in Canary M112.
- See diff between `Chrome111` and `Chrome74` here: https://www.diffchecker.com/tIkXlhzv/

### Some references:

- Available `scalabilityMode` in libwebrtc for VP8: https://webrtc.googlesource.com/src/+/a1a7c638ece600a0d9e2a1f7d77e1a9050737e90/modules/video_coding/codecs/vp8/vp8_scalability.h
- Available `scalabilityMode` in libwebrtc for VP9 and H264: https://webrtc.googlesource.com/src/+/ecfe8da46b22883363a66d0909d7d2246cb13c48/modules/video_coding/svc/create_scalability_structure.cc
- See new query params in mediasoup-demo: https://github.com/versatica/mediasoup-demo/#configuration-via-query-parameters
- Thanks @thebongy for documenting this in https://github.com/versatica/mediasoup-client/pull/247. 

### Issues

_NOTE:_ These issues are not necessarily intended for this PR.

**ISSUE 1:** If the user creates a VP8 producer with  3 encodings but do not include `scalabilityMode` into them, then Chrome will select 'L3T3' but we will signal "1 spatial and 1 temporal layer" to mediasoup, so consumers will just receive the temporal layer 0 of each simulcast stream.
  - **UPDATE:** Fixed in https://github.com/versatica/mediasoup-client/pull/248/commits/ae22bd9c4c222238eb931b66732707ce5c67b50c

**ISSUE 2:** We must allow creation of a `SimulcastConsumer` in mediasoup side (instead of `SingleConsumer`) when the producer `RtpParameters` have a single `encoding` with `scalabilityMode` other than 'L1T1'.
  - Of course, in the future we should just merge all `XxxxxConsumer` classes into one as discussed some time ago.
  - **UPDATE:** PR done in mediasoup: https://github.com/versatica/mediasoup/pull/1013

**ISSUE 3:** User may pass different `scalabilityMode` for each encoding in simulcast (it works). Problem is that in mediasoup land we assume that number of temporal layers is the same for all the simulcast streams. In fact AFAIR we read it from the first encoding.
  - **UPDATE:** Fixed here: https://github.com/versatica/mediasoup-client/pull/248/commits/b9435d1f8b383559093ec96b027f9f07b5fe1670

**ISSUE 4:** There are no temporal layers when using H264 simulcast, but it doesn't even work using `Chrome74` handler in Chrome or Canary old or latest versions. ¯\_(ツ)_/¯

**ISSUE 5:** Screen sharing with VP9 (with or without SVC) does not work in Canary M111. No idea. It wouldn't be the first time something similar happens in Canary last version.

### DIFF
--- a/src/Device.ts
+++ b/src/Device.ts
@@ -281,7 +281,7 @@ export class Device
 
 			switch (handlerName)
 			{
-			case 'Chrome111':
+				case 'Chrome111':
 					this._handlerFactory = Chrome111.createFactory();
 					break;
 				case 'Chrome74':
--- a/src/Device.ts
+++ b/src/Device.ts
@@ -6,6 +6,7 @@ import * as utils from './utils';
 import * as ortc from './ortc';
 import { Transport, TransportOptions, CanProduceByKind } from './Transport';
 import { HandlerFactory, HandlerInterface } from './handlers/HandlerInterface';
+import { Chrome111 } from './handlers/Chrome111';
 import { Chrome74 } from './handlers/Chrome74';
 import { Chrome70 } from './handlers/Chrome70';
 import { Chrome67 } from './handlers/Chrome67';
@@ -22,6 +23,7 @@ import { SctpCapabilities } from './SctpParameters';
 const logger = new Logger('Device');
 
 export type BuiltinHandlerName =
+	| 'Chrome111'
 	| 'Chrome74'
 	| 'Chrome70'
 	| 'Chrome67'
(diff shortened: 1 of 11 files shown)


jmillan (approved):
:+1:

ibc:
Released in mediasoup-client 3.6.80

Rush:
> **ISSUE 5:** Screen sharing with VP9 (with our without SVC) does not work in Canary M111. No idea. It wouldn't be the first time something similar happens in Canary last version.

Did anybody figure out a fix? It seems it does work if I do not include any scalabilityMode but how to make it work with scalabilityMode?

ibc:
> Did anybody figure out a fix? It seems it does work if I do not include any scalabilityMode but how to make it work with scalabilityMode?

Please report the issue in libwebrtc project if you wish. Let's not use this PR as a forum.

PR #250, opened 2023-02-27
implement replaceTrack in ReactNative

okamoto00:
`replaceTrack()` was not implemented in ReactNative, so it added.


ibc (commented):
This PR is not replacing any sending track but just replacing a track in an internal map. It doesn't do anything. The limitation is not artificial. It just happens that the version of react-native-webrtc in ReactNative handler does NOT support replaceTrack() so it cannot be added.

Anyway you should use the new ReactNativeUnifiedPlan handler which supports replaceTrack().

Closing.

PR #257, opened 2023-04-07
Make AppData typed and writable

ibc:
Same as done in mediasoup-node.

### Usage

```ts
const audioProducer = await sendTransport.produce<{ foo: string }>(
	{
		track: audioTrack,
		stopTracks: false,
		appData: { foo: 'FOO' }
	});
```

(same in all `device.createSend/RecvTransport()`, `transport.consume()`, `transport.produceData()` and `transport.consumeData()`.

### DIFF
--- a/src/tests/test.ts
+++ b/src/tests/test.ts
@@ -185,15 +185,20 @@ test('device.createSendTransport() for sending media succeeds', () =>
 		sctpParameters
 	} = fakeParameters.generateTransportRemoteParameters();
 
-	sendTransport = device.createSendTransport(
+	sendTransport = device.createSendTransport<{ foo: number }>(
 		{
 			id,
 			iceParameters,
 			iceCandidates,
 			dtlsParameters,
 			sctpParameters,
-			appData : { baz: 'BAZ' }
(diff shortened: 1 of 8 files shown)


src/tests/test.ts
```diff
@@ -185,24 +185,29 @@ test('device.createSendTransport() for sending media succeeds', () =>
 		sctpParameters
 	} = fakeParameters.generateTransportRemoteParameters();
 
-	sendTransport = device.createSendTransport(
+	sendTransport = device.createSendTransport<{ foo: number }>(
```
dlmanning, line 188: Your issue is that above you've already defined `sendTransport` as a `mediasoupClient.types.Transport`, which has a default type for `AppData` of `{ [key: string]: unknown }`

You can see what I mean just by putting a `const` at the beginning of line 188 so the compiler sees this as a fresh binding.

ibc: OMG so you mean that it's just an issue in the tests file because I'm reusing a let variable???

ibc:
So thanks @dlmanning.

PR #260, opened 2023-05-17
Create ESModule bundle by vite

satoren:
The use of ESModule allows tree-shaking to work in many bundlers, reducing bundle size slightly.

### DIFF
--- a/npm-scripts.js
+++ b/npm-scripts.js
@@ -167,7 +167,7 @@ function lint()
 {
 	console.log('npm-scripts.js [INFO] lint()');
 
-	executeCmd('eslint -c .eslintrc.js --max-warnings 0 src .eslintrc.js npm-scripts.js');
+	executeCmd('eslint -c .eslintrc.js --max-warnings 0 src .eslintrc.js npm-scripts.js vite.config.js');
 }
 
 function test()
--- a/npm-scripts.js
+++ b/npm-scripts.js
@@ -160,6 +160,7 @@ function buildTypescript(force = false)
 	deleteLib();
 
 	executeCmd('tsc');
+	executeCmd('vite build');
 }
 
 function lint()
--- a/vite.config.js
+++ b/vite.config.js
@@ -1,0 +1,21 @@
+import { resolve } from 'path';
+import { defineConfig } from 'vite';
+import typescript from '@rollup/plugin-typescript';
+
+export default defineConfig({
+	build : {
+		lib : {
+			// Could also be a dictionary or array of multiple entry points
+			entry    : resolve(__dirname, 'src/index.ts'),
+			name     : 'mediasoup-client',
+			fileName : 'index',
+			formats  : [ 'cjs', 'es' ]      
+		},
+		rollupOptions : {
+			plugins : [ typescript() ],
+			output  : {
+				dir : 'lib'
+			}
+		}
+	}
+});
--- a/vite.config.js
+++ b/vite.config.js
@@ -5,11 +5,10 @@ import typescript from '@rollup/plugin-typescript';
 export default defineConfig({
 	build : {
 		lib : {
-			// Could also be a dictionary or array of multiple entry points
 			entry    : resolve(__dirname, 'src/index.ts'),
 			name     : 'mediasoup-client',
 			fileName : 'index',
-			formats  : [ 'cjs', 'es' ]      
+			formats  : [ 'cjs', 'es' ]
 		},
 		rollupOptions : {
 			plugins : [ typescript() ],
--- a/vite.config.js
+++ b/vite.config.js
@@ -2,19 +2,25 @@ import { resolve } from 'path';
 import { defineConfig } from 'vite';
 import typescript from '@rollup/plugin-typescript';
 
-export default defineConfig({
-	build : {
-		lib : {
-			entry    : resolve(__dirname, 'src/index.ts'),
-			name     : 'mediasoup-client',
-			fileName : 'index',
-			formats  : [ 'cjs', 'es' ]
-		},
-		rollupOptions : {
-			plugins : [ typescript() ],
-			output  : {
-				dir : 'lib'
+export default defineConfig(
+	{
+		build :
+		{
+			lib :
+			{
+				// Could also be a dictionary or array of multiple entry points.
+				entry    : resolve(__dirname, 'src/index.ts'),
+				name     : 'mediasoup-client',
+				fileName : 'index',
+				formats  : [ 'cjs', 'es' ]
+			},
+			rollupOptions :
+			{
+				plugins : [ typescript() ],
+				output  :
+				{
+					dir : 'lib'
+				}
 			}
 		}
-	}
-});
+	});
--- a/package-lock.json
+++ b/package-lock.json
@@ -21,6 +21,7 @@
         "supports-color": "^9.3.1"
       },
       "devDependencies": {
+        "@rollup/plugin-typescript": "^11.1.1",
         "@types/events": "^3.0.0",
         "@types/jest": "^29.5.1",
         "@types/sdp-transform": "^2.4.6",
@@ -32,7 +33,8 @@
         "open-cli": "^7.2.0",
         "ts-jest": "^29.1.0",
         "tsc-watch": "^6.0.4",
-        "typescript": "^5.0.4"
+        "typescript": "^5.0.4",
+        "vite": "^4.3.7"
       },
       "engines": {
         "node": ">=14"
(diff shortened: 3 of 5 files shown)


npm-scripts.js
```diff
@@ -160,6 +160,7 @@ function buildTypescript(force = false)
 	deleteLib();
 
 	executeCmd('tsc');
+	executeCmd('vite build');
```
ibc, line 163: Why should we run this every time we generate JS code? Why here?

ibc:
> The use of ESModule allows tree-shaking to work in many bundlers, reducing bundle size slightly.

I have no idea what this is, honestly.

Why should we use "exnest" in tsconfig.json which cane output something completely different depending when it's executed?

Also, mediasoup-client is supposed to also work in Node where import/export may not be available depending on Node.js version. Is this change breaking it by changing CommonJS to Module?

BTW do you know that the mediasoup-client package only contains JS code and it doesn't contain TypeScript code? So how is this thing useful at all? What does "reducing bundle size slightly" mean exactly? Which bundle?

vite.config.js
ibc, line 1: The "lint" script in package.json must be updated to also check this file since syntax in this file is wrong:

- Missing blank line at the end.
- Comments must start with capital letter and end with dot.
- Brackets must be in next line.
- etc.

ibc:
Also this PR is including vite dep which is 3.25 MB and includes things about frontend, static server and CSS that are completely unnecessary for mediasoup-client.

npm-scripts.js
satoren: Generate js code with vite. [tsc is used only for type checking](https://github.com/versatica/mediasoup-client/pull/260/files#diff-b55cdbef4907b7045f32cc5360d48d262cca5f94062e353089f189f4460039e0R15).

satoren:
> Why should we use "exnest" in tsconfig.json which cane output something completely different depending when it's executed?

Changed because vite needed to change from commonjs to reference ES2015.
Since we are not using tsc to create the js code, the output code will probably not change.


> Also, mediasoup-client is supposed to also work in Node where import/export may not be available depending on Node.js version. Is this change breaking it by changing CommonJS to Module?

It is set to output both common js and esmodule. index.js is commonjs index.mjs is esmodule.


> BTW do you know that the mediasoup-client package only contains JS code and it doesn't contain TypeScript code? So how is this thing useful at all? What does "reducing bundle size slightly" mean exactly? Which bundle?

Have you ever heard of tree-shaking, a bundler function that removes unused code from an imported file?

satoren:
> Also this PR is including vite dep which is 3.25 MB and includes things about frontend, static server and CSS that are completely unnecessary for mediasoup-client.

Is it so important to use a few MB of devDependency package? I think it is more important that the final product be a few KB less.

satoren:
For example, in the esModule version of lodash, only 4.4 kB will be included in the final code if you only want to use isEqual. (The commonjs version adds 24.5 kB)

https://bundlephobia.com/package/lodash@4.17.21
https://bundlephobia.com/package/lodash-es@4.17.21

The effect on the mediasoup-client is limited because tree-shaking does not actually work on a per-class-method basis, but the handler code is large and may be useful for optimization there.

npm-scripts.js
ibc: Is that the `noEmit: true` option in `tsconfig.json`?

ibc:
> Is it so important to use a few MB of devDependency package? I think it is more important that the final product be a few KB less.

Definitely the size of the dev dependencies is not that important. BTW this is the generated JS size reduction:

```
$ du -sh lib
1.0M	lib

$ du -sh lib-esmodule/
944K	lib-esmodule/
```

Is it that much? How is it relevant?

vite.config.js
ibc: Can you please add this commit into your branch? I don't have permission to push to it directly:

https://github.com/versatica/mediasoup-client/commit/811c6d9abc67725bfa5865ef9817227f6d5c1c0f

ibc:
So I cannot use this mediasoup-client branch in the mediasoup-demo because `gulp` in `mediasoup-demo/app` folder now fails:

```
SyntaxError: Unexpected token (6:25116) while parsing /Users/ibc/src/v3-mediasoup-demo/app/node_modules/mediasoup-client/lib/index.js while parsing file: /Users/ibc/src/v3-mediasoup-demo/app/node_modules/mediasoup-client/lib/index.js
    at DestroyableTransform.end [as _flush] (/Users/ibc/src/v3-mediasoup-demo/app/node_modules/insert-module-globals/index.js:114:21)
    at DestroyableTransform.prefinish (/Users/ibc/src/v3-mediasoup-demo/app/node_modules/readable-stream/lib/_stream_transform.js:138:10)
    at DestroyableTransform.emit (node:events:527:28)
    at DestroyableTransform.emit (node:domain:537:15)
    at prefinish (/Users/ibc/src/v3-mediasoup-demo/app/node_modules/readable-stream/lib/_stream_writable.js:619:14)
    at finishMaybe (/Users/ibc/src/v3-mediasoup-demo/app/node_modules/readable-stream/lib/_stream_writable.js:627:5)
    at endWritable (/Users/ibc/src/v3-mediasoup-demo/app/node_modules/readable-stream/lib/_stream_writable.js:638:3)
    at DestroyableTransform.Writable.end (/Users/ibc/src/v3-mediasoup-demo/app/node_modules/readable-stream/lib/_stream_writable.js:594:41)
    at DestroyableTransform.onend (/Users/ibc/src/v3-mediasoup-demo/app/node_modules/readable-stream/lib/_stream_readable.js:577:10)
    at Object.onceWrapper (node:events:641:28)
```

ibc:
> So I cannot use this mediasoup-client branch in the mediasoup-demo because `gulp` in `mediasoup-demo/app` folder now fails:
> 
> ```
> SyntaxError: Unexpected token (6:25116) while parsing /Users/ibc/src/v3-mediasoup-demo/app/node_modules/mediasoup-client/lib/index.js while parsing file: /Users/ibc/src/v3-mediasoup-demo/app/node_modules/mediasoup-client/lib/index.js
>     at DestroyableTransform.end [as _flush] (/Users/ibc/src/v3-mediasoup-demo/app/node_modules/insert-module-globals/index.js:114:21)
>     at DestroyableTransform.prefinish (/Users/ibc/src/v3-mediasoup-demo/app/node_modules/readable-stream/lib/_stream_transform.js:138:10)
>     at DestroyableTransform.emit (node:events:527:28)
>     at DestroyableTransform.emit (node:domain:537:15)
>     at prefinish (/Users/ibc/src/v3-mediasoup-demo/app/node_modules/readable-stream/lib/_stream_writable.js:619:14)
>     at finishMaybe (/Users/ibc/src/v3-mediasoup-demo/app/node_modules/readable-stream/lib/_stream_writable.js:627:5)
>     at endWritable (/Users/ibc/src/v3-mediasoup-demo/app/node_modules/readable-stream/lib/_stream_writable.js:638:3)
>     at DestroyableTransform.Writable.end (/Users/ibc/src/v3-mediasoup-demo/app/node_modules/readable-stream/lib/_stream_writable.js:594:41)
>     at DestroyableTransform.onend (/Users/ibc/src/v3-mediasoup-demo/app/node_modules/readable-stream/lib/_stream_readable.js:577:10)
>     at Object.onceWrapper (node:events:641:28)
> ```

Probably related to things like this in the bundled `index.js` (not really sure):

```
{super(e??"AwaitQueue task removed")
```

satoren:
The gulp command passes in my environment. I wonder what's the difference. 🤔 
```
 app % gulp
[09:17:46] NODE_ENV: development
[09:17:46] Using gulpfile ~/mediasoup-demo/app/gulpfile.js
[09:17:46] Starting 'default'...
[09:17:46] Starting 'dist'...
[09:17:46] Starting 'clean'...
[09:17:46] Finished 'clean' after 17 ms
[09:17:46] Starting 'lint'...
[09:17:46] 
/Users/masataka/mediasoup-demo/app/gulpfile.js
  58:21  error  Unexpected use of file extension "json" for "./package.json"  import/extensions

/Users/masataka/mediasoup-demo/app/lib/components/PeerView.jsx
  455:6  error  Invalid property 'playsInline' found on tag 'audio', but it is only allowed on: video  react/no-unknown-property

✖ 2 problems (2 errors, 0 warnings)

[09:17:46] Finished 'lint' after 916 ms
[09:17:46] Starting 'bundle'...
[09:17:51] Finished 'bundle' after 4.31 s
[09:17:51] Starting 'html'...
[09:17:51] Finished 'html' after 6.48 ms
[09:17:51] Starting 'css'...
[09:17:51] Finished 'css' after 283 ms
[09:17:51] Starting 'resources'...
[09:17:51] Finished 'resources' after 26 ms
[09:17:51] Finished 'dist' after 5.56 s
[09:17:51] Finished 'default' after 5.56 s
```

npm-scripts.js
satoren: Yes, it is.

palashgo:
> > Is it so important to use a few MB of devDependency package? I think it is more important that the final product be a few KB less.
> 
> Definitely the size of the dev dependencies is not that important. BTW this is the generated JS size reduction:
> 
> ```
> $ du -sh lib
> 1.0M	lib
> 
> $ du -sh lib-esmodule/
> 944K	lib-esmodule/
> ```
> 
> Is it that much? How is it relevant?

The original `lib` output directory consists of types and `cjs` js modules. 
This contains two bundles `cjs` and `esm`

Also I believe the PR misses this step of excluding public dependencies, the current build inlines all external dependencies unless you explicitly mark them external. Applying the below changes gives a more direct comparison vs the current tsc build step which also does not inline your dependencies

```
diff --git a/vite.config.js b/vite.config.js
index a258f68..34c2a41 100644
--- a/vite.config.js
+++ b/vite.config.js
@@ -1,6 +1,7 @@
 import { resolve } from 'path';
 import { defineConfig } from 'vite';
 import typescript from '@rollup/plugin-typescript';
+import pkg from './package.json';
 
 export default defineConfig(
 	{
@@ -17,6 +18,7 @@ export default defineConfig(
 			rollupOptions :
 			{
 				plugins : [ typescript() ],
+				external : Object.keys(pkg.dependencies),
 				output  :
 				{
 					dir : 'lib'

```

With this you get

```
du -sh lib/
816K	lib/
```

Although I believe all these numbers are meaningless because currently `tsc` just transpiles Typescript to Javascript, `vite` has a bundler step which bundles everything into a single file and these optimisations are from bundling and not from tree shaking. When a client integrates mediasoup-client library they will have a bundler + minifier steps in their build which will apply these optimisations anyway.
What should be measured is the difference before and after adding mediasoup libraries to clients under different build tooling (rollup/esbuild/webpack/bun)

ES bundles in general improve tree shaking but there is very little in the current build which is tree shakable so the benefits right now are little to negligible 

--------

Irrelevant to the PR but relevant to the discussions about reducing the size impact of mediasoup-client, we at dyte.io 

1. Split the file `Device.ts` into a base class with abstract `switchDevice` methods implemented in `FullDevice.ts` which contains all the handlers and `BrowserDevice.ts` which contains only browser handlers (No React Native handlers)

Now using the package.json ["browser" field](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#browser) we can swap out FullDevice with BrowserDevice on builds targeting browser

```
"browser": {
    "lib/FullDevice.js":"lib/BrowserDevice.js"
  }
  ```
  
2. Majority of the handler code is duplicated, eg Chrome74 vs Chrome111 have just a few lines of differences, we have tried reducing the number of handlers and incorporating the differences inside the relevant methods directly


Happy to create a PR to this upstream repo for both of these things

satoren:
@palashgo Thanks for the detailed discussion.
> Also I believe the PR misses this step of excluding public dependencies, the current build inlines all external dependencies unless you explicitly mark them external. Applying the below changes gives a more direct comparison vs the current tsc build step which also does not inline your dependencies

I had forgotten that perspective. Whether to inline or not needs to be verified separately.

> Although I believe all these numbers are meaningless as when a client integrates mediasoup-client library they will have a bundler + minifier steps in their build which will apply these optimisations anyway.
What should be measured is the difference before and after adding mediasoup libraries to clients under different build tooling (rollup/esbuild/webpack/bun)

> ES bundles in general improve tree shaking but there is nothing in the current build which is tree shakable so the benefits right now are little to negligible

I agree that the output size as a library is almost meaningless. Therefore, I compared the final product code with the branch to which this PR was applied and found the following reductions(using vite rollup )
* before
`dist/assets/index-aa6ffcf2.js                                  434.98 kB │ gzip:  92.95 kB`
* after
`dist/assets/index-eb2715f9.js                                  402.86 kB │ gzip:  87.34 kB`

This may not be because of the esmodule, but because the code converted to commonjs by tsc may be hard for the minifier to work.

I was thinking of the idea of allowing the user to select which handlers to enable and reduce it by tree-shaking. (Our product does not support too old browsers such as Chrome 55, Edge11 or Safari 11). 
But your idea sounds good too.

satoren:
When I tried it with another vite project(bundler rollup), I found the following differences.

The differences between cjs and esm were small, but the differences between them and tsc were significant.

* vite commonjs(using index.js in this branch) `dist/assets/main.3dbbee19.js   328.08 KiB / gzip: 66.27 KiB`
* vite esmodule(using index.mjs in this branch) `dist/assets/main.fd5c54fa.js   327.33 KiB / gzip: 65.74 KiB`
* tsc (current master) `dist/assets/main.99801bb4.js   357.59 KiB / gzip: 70.79 KiB`

palashgo:
> The differences between cjs and esm were small, but the differences between them and tsc were significant.


@satoren Interesting, let me try this in our builds

satoren:
https://github.com/versatica/mediasoup-client/pull/267
https://github.com/versatica/mediasoup-client/pull/268

PR #266, opened 2023-06-15
Replace bowser with ua-parser-js

### LINKED ISSUE #262: Replace bowser dep
### Your environment

- Operating system: Android
- Browser version: Chrome 104
- npm version: *
- mediasoup version: *
- mediasoup-client version: *

### Issue description

mediasoup-client uses `bowser` for browser detection and for switching to appropriate handlers.
[Bowser](https://www.npmjs.com/package/bowser) however is an unmaintained project, last update was published 3 years ago and there are tons of open issues

Example of one issue we faced in production is with this user agent (Chrome Android, Motorola Edge)

```
Mozilla/5.0 (Linux; Android 10; motorola edge) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/104.0.0.0 Mobile Safari/537.36
```

This is how `bowser` parses this user agent

```
browser name: "Microsoft Edge"
os name: "Android"
version: "10"
engine name: "EdgeHTML"
version: ""
```

Which leads mediasoup-client to incorrectly use the Edge handlers instead of the Chrome handlers

One way to solve this would be to switch to a more modern and better supported user agent parsing library - [ua-parser-js](https://www.npmjs.com/package/ua-parser-js)

```
browser name: "Mobile Chrome"
os name: "Android"
version: "10"
engine name: "Blink"
version: "104.0.0.0"
```

Both these projects includes complete test suites, so a switch to ua-parser-js could be made asserting bowser tests so nothing that was working before would break

Another alternative is to switch to a feature based browser detection which would be more reliable and would solve for issues with browser wrappers and in app browsers

ibc:
Fixes #262

### DIFF
--- a/src/Device.ts
+++ b/src/Device.ts
@@ -1,4 +1,4 @@
-import Bowser from 'bowser';
+import { UAParser } from 'ua-parser-js';
 import { Logger } from './Logger';
 import { EnhancedEventEmitter } from './EnhancedEventEmitter';
 import { UnsupportedError, InvalidStateError } from './errors';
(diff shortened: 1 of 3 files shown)


src/Device.ts
```diff
@@ -85,64 +87,79 @@ export function detectDevice(): BuiltinHandlerName | undefined
...
 		// Firefox.
-		else if (browser.satisfies({ firefox: '>=60' }))
+		else if (browserName === 'firefox' && osName !== 'ios' && browserVersion >= 60)
 		{
 			return 'Firefox60';
 		}
-		// Firefox on iOS.
-		else if (browser.satisfies({ ios: { OS: '>=14.3', firefox: '>=30.0' } }))
+		// Firefox on iOS (so Safari).
+		else if (browserName === 'firefox' && osName === 'ios' && osVersion >= 14.3)
```
palashgo, line 137: Is the firefox version check `firefox: '>=30.0'` drop intended

jmillan (approved):

src/Device.ts
ibc: I don't remember what such a "Firefox version 30" was honestly. Are you requesting this change based on previous code? Or because you know it makes sense? The thing is: if we are checking that iOS version is >= 14.3 then why do we need to also  check Firefox version? We know it's Safari anyway and we know is a modern version of it.

ibc: I'm gonna merge and release. If you still think this is a required change, please comment later.

palashgo: Yeah I just wanted to make sure if the drop was intentional, I don't see a reason for including a firefox version as well

palashgo:
@ibc While running some tests, noticed something critical `ua-parser-js` returns "Mobile Chrome", "Mobile Safari", "Mobile Firefox" instead of "Chrome", "Safari" and "Firefox" 
This would then break in all non Mobile Safari and Firefox devices ( Chromium ones might still be handled by the "Best effort for Chromium based browsers." case which checks the engine name )

PR #267, opened 2023-06-16
Build esm/commonjs using rollup

satoren:
Brushed up https://github.com/versatica/mediasoup-client/pull/260.

There are no more extras than with vite.

PR #268, opened 2023-06-16
Selective handler for Device

satoren:
See https://github.com/versatica/mediasoup-client/pull/267 first  to add esm build. 

Allow flexible selection of handlers as follows

```tsx
import { ManualDevice, detectDevice } from "mediasoup-client";
import {
  Chrome111,
  Chrome74,
  Chrome70,
  Firefox60,
  Safari12,
} from "mediasoup-client/builtin-handlers";

const handlerFactory = () => {
  switch (detectDevice()) {
    case "Chrome111":
      return Chrome111.createFactory();
    case "Chrome74":
      return Chrome74.createFactory();
    case "Chrome70":
      return Chrome70.createFactory();
    case "Firefox60":
      return Firefox60.createFactory();
    case "Safari12":
      return Safari12.createFactory();
    default:
      return Chrome70.createFactory();
  }
};

const device = new ManualDevice({ handlerFactory: handlerFactory() });

```

Combined with ESM builds, tree shaking can reduce the bundle size by more than 160 kb.

For compatibility, I created a `ManualDevice` from `Device` with `handlerFactory` as a required parameter.

### DIFF
--- a/npm-scripts.mjs
+++ b/npm-scripts.mjs
@@ -164,6 +164,7 @@ function buildTypescript(force = false)
 
 	deleteLib();
 	executeCmd('tsc');
+	executeCmd('rollup --config');
 }
 
 function lint()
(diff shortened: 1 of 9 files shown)


ibc:
@satoren, thanks for this PR. Although it's not valid for v3 (due to given rationale above) I'll consider it for next major version.

PR #275, opened 2023-07-13
Handlers: check if the handler is closed and call handler methods with await

### LINKED ISSUE #274: producer.close() throws unhandled rejection if called when PeerConnection's signalingState is "closed"
CleanShot-2023-07-13-at-14 24 17@2x

So if `producer.close()` has been called and later `Transport` calls `handler.stopSending()` this happens in `Chrome111` (and probably in many other handlers):

```ts
async stopSending(localId: string): Promise<void>
{
		this.assertSendDirection();

		logger.debug('stopSending() [localId:%s]', localId);

		const transceiver = this._mapMidTransceiver.get(localId);

		if (!transceiver)
		{
			throw new Error('associated RTCRtpTransceiver not found');
		}

		transceiver.sender.replaceTrack(null);

		this._pc.removeTrack(transceiver.sender);

		const mediaSectionClosed =
			this._remoteSdp!.closeMediaSection(transceiver.mid!);
```

and `this._pc.removeTrack(transceiver.sender);` will throw with "DOMException: The peer connection is closed".

It happens here in `Transport.ts`:
CleanShot-2023-07-13-at-14 27 35@2x

This block is just wrong:

```ts
this._awaitQueue.push(
				async () => this._handler.stopSending(producer.localId),
				'producer @close event')
				.catch((error: Error) => logger.warn('producer.close() failed:%o', error));
```

Here we are NOT doing `await this._handler.stopSending(producer.localId)` or `this._handler.stopSending(producer.localId).catch(() => {})` so if it rejects then an unhandled exception happens.

So possible solutions:

### Solution 1

In all methods in handlers (specially in `stopSending()` check if `this._pc.signalingState === 'closed'` and if so do nothing (early return).

### Solution 2

In the `producer.on('@close')` listener in `Transport.ts` add `await` to `this._handler.stopSending(producer.localId)`. Problems:
1. It will make things slower since the queue must wait for the `stopSending()` method to resolve (or reject) before it processes next enqueued task.
2. It will still show the warning "producer.close() failed". We don't want to show that error since it's not legitimate. We closed the Producer already so it shouldn't happen.

### Proposed Solution

A mix between both above:

- Check if `this._pc.signalingState === 'closed'` in all handler methods and return early if so. NOTE: Not in all methods because ofc we want that `producer.pause()` throws if the Producer is closed.
- Add `await` into all handler related tasks passed to the queue in `Transport.ts`.

### WOW!

But this is funny because **right now the queue is NOT enqueing anything**. It's just enqueing NON async methods since there is no `await` at all. To be clear, `this._awaitQueue.push(async () => this._handler.stopSending(producer.localId) ...` is like introducing a sync function in the queue. The `task` given to `AwaitQueue` is `async () => this._handler.stopSending(producer.localId)`. This is gonna return **immediately**. For it to really wait until the task completes, it should be `async () => await this._handler.stopSending(producer.localId)`.

ibc:
**NOTE:** Draft PR for now since chnages are only done in `Chrome111` handler.

- Fixes #274
- Add "closed" flag in handlers.
- Make some handler methods throw if closed.
- Make some other methods return early (without doing anything) if the handler is closed. Yes, methods that close or stop things should be idempotent and never throw.
- `Transport.ts`: use `await` in all calls to handler methods given to the `AwaitQueue`.

### DIFF
--- a/src/Transport.ts
+++ b/src/Transport.ts
@@ -449,7 +449,7 @@ export class Transport<TransportAppData extends AppData = AppData>
 
 		// Enqueue command.
 		return this._awaitQueue.push(
-			async () => this._handler.restartIce(iceParameters),
+			async () => await this._handler.restartIce(iceParameters),
 			'transport.restartIce()');
 	}
 
@@ -1247,7 +1247,7 @@ export class Transport<TransportAppData extends AppData = AppData>
 			}
 
 			this._awaitQueue.push(
-				async () => this._handler.stopSending(producer.localId),
+				async () => await this._handler.stopSending(producer.localId),
 				'producer @close event')
 				.catch((error: Error) => logger.warn('producer.close() failed:%o', error));
 		});
@@ -1255,7 +1255,7 @@ export class Transport<TransportAppData extends AppData = AppData>
 		producer.on('@pause', (callback, errback) =>
 		{
 			this._awaitQueue.push(
-				async () => this._handler.pauseSending(producer.localId),
+				async () => await this._handler.pauseSending(producer.localId),
 				'producer @pause event')
 				.then(callback)
 				.catch(errback);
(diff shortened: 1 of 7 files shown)


jmillan (approved):

ibc:
I've completed other handlers. Not all but just the latest versions. Honestly having to maintain so many old handlers is hell and with the changes in `Transport` (the addition of those `await`) the bug is also fixed for all handlers (it's just that some of them may show a warning log, nothing else).

PR #277, opened 2023-08-19
Fix ufrag is not required at media level

ggarber:
Extracting DTLS parameters fails when the iceUFrag/pwd is defined at session level instead of the media level.

### DIFF
--- a/src/handlers/sdp/commonUtils.ts
+++ b/src/handlers/sdp/commonUtils.ts
@@ -172,8 +172,8 @@ export function extractDtlsParameters(
 ): DtlsParameters
 {
 	const mediaObject = (sdpObject.media || [])
-		.find((m: { iceUfrag: string; port: number }) => (
-			m.iceUfrag && m.port !== 0
+		.find((m: { port: number }) => (
+			m.port !== 0
 		));
 
 	if (!mediaObject)


src/handlers/sdp/commonUtils.ts
```diff
@@ -172,8 +172,8 @@ export function extractDtlsParameters(
 ): DtlsParameters
 {
 	const mediaObject = (sdpObject.media || [])
-		.find((m: { iceUfrag: string; port: number }) => (
-			m.iceUfrag && m.port !== 0
+		.find((m: { port: number }) => (
```
ibc, line 175: I think we should at least verify that the media section has `a=setup` attribute, right? Note that in lines below we return a value based on `mediaObject.setup` value.

ibc: I'm gonna merge and then I'll add that check. Thanks!

ibc: Done here: https://github.com/versatica/mediasoup-client/commit/eb90e61cbd1da810a5ac271fed02d650706802e6

ibc:
Released in 3.6.99

src/handlers/sdp/commonUtils.ts
ggarber: Thank you @ibc 

Not sure if there is anything more updated but 4145 says that "The 'setup' attribute is charset-independent and can be a session-level or a media-level attribute."   so in theory you could have a similar problem.

I just checked it against the SDP I was having problems with (It is actually OBS WHIP) and setup is actually sent at session level too :(.  So my suggestion is to revert [eb90e61](https://github.com/versatica/mediasoup-client/commit/eb90e61cbd1da810a5ac271fed02d650706802e6)

ibc: Oh... ok but it's not just about reverting it. We must also be ready to read a=setup from session level too. Could you please fill an issue? I'd do but I must go to Aste Nagusia soon.

ibc: I've properly fixed it in this commit: https://github.com/versatica/mediasoup-client/commit/0381230b93abb1a0b296c21d1361037be552e870

ibc: Released in 3.6.100

PR #280, opened 2023-10-05
Detect embedded webview on iPadOS.

bclark-videra:
iPadOS identifies as Mac OS, but ua-parser-js does the extra detection to determine if it's an iPad. Use the device model to detect when we're in an embedded webview on iPadOS.

Also removes the version check for best effort webkit detection as iPadOS sends version 10.15 and checking RTCRtpTransceiver should be sufficient without the version check.

### DIFF
--- a/src/Device.ts
+++ b/src/Device.ts
@@ -106,7 +106,7 @@ export function detectDevice(): BuiltinHandlerName | undefined
 		const device = uaParser.getDevice();
 		const deviceModel = device.model?.toLowerCase() ?? '';
 
-		const isIOS = osName === 'ios' || (osName === 'mac os' && deviceModel === 'ipad');
+		const isIOS = osName === 'ios' || deviceModel === 'ipad';
 
 		const isChrome =
 		[
--- a/src/Device.ts
+++ b/src/Device.ts
@@ -103,8 +103,10 @@ export function detectDevice(): BuiltinHandlerName | undefined
 		const os = uaParser.getOS();
 		const osName = os.name?.toLowerCase() ?? '';
 		const osVersion = parseFloat(os.version ?? '0');
+		const device = uaParser.getDevice();
+		const deviceModel = device.model?.toLowerCase() ?? '';
 
-		const isIOS = osName === 'ios';
+		const isIOS = osName === 'ios' || (osName === 'mac os' && deviceModel === 'ipad');
 
 		const isChrome =
 		[
@@ -188,7 +190,6 @@ export function detectDevice(): BuiltinHandlerName | undefined
 		else if (
 			engineName === 'webkit' &&
 			isIOS &&
-			osVersion >= 14.3 &&
 			typeof RTCRtpTransceiver !== 'undefined' &&
 			RTCRtpTransceiver.prototype.hasOwnProperty('currentDirection'))
 		{


bclark-videra:
Debug log from device detection from an embedded WKWebView on iPadOS without this change:
```
[Info] mediasoup-client:Device constructor() +0ms
[Info] mediasoup-client:Device detectDevice() | browser detected [ua:Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_6) AppleWebKit/605.1.15 (KHTML, like Gecko), parsed:
{ua: "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_6) AppleWebKit/605.1.15 (KHTML, like Gecko)", browser: {name: "WebKit", version: "605.1.15", major: "605"}, engine: {name: "WebKit", version: "605.1.15"}, os: {name: "Mac OS", version: "10.15.6"}, device: {vendor: "Apple", model: "iPad", type: "tablet"}, …}
] +3ms
[Warning] mediasoup-client:WARN:Device detectDevice() | browser not supported [name:webkit, version:605] +0ms
```

src/Device.ts
```diff
@@ -103,8 +103,10 @@ export function detectDevice(): BuiltinHandlerName | undefined
 		const os = uaParser.getOS();
 		const osName = os.name?.toLowerCase() ?? '';
 		const osVersion = parseFloat(os.version ?? '0');
+		const device = uaParser.getDevice();
+		const deviceModel = device.model?.toLowerCase() ?? '';
 
-		const isIOS = osName === 'ios';
+		const isIOS = osName === 'ios' || (osName === 'mac os' && deviceModel === 'ipad');
```
ibc: What about just this?

```suggestion
		const isIOS = osName === 'ios' || deviceModel === 'ipad';
```

src/Device.ts
```diff
@@ -188,7 +190,6 @@ export function detectDevice(): BuiltinHandlerName | undefined
 		else if (
 			engineName === 'webkit' &&
 			isIOS &&
-			osVersion >= 14.3 &&
```
ibc, line 191: I understand this is not super required, but what is wrong with it? Is this version requirement generating some issue?

bclark-videra: Yes, iPadOS WKWebView reports as OS Version 10.15.

ibc:
Thanks, I'll merge and release next Monday or before if I can.

ibc:
Sorry for the delay. Released in 3.6.102.

bclark-videra:
> Sorry for the delay. Released in 3.6.102.

Thank you!

iamarkdev:
This issue appears to be back for the latest iPad OS.

ibc:
> This issue appears to be back for the latest iPad OS.

Having the UserAgent string of latest iPad OS would help.

iamarkdev:
> > This issue appears to be back for the latest iPad OS.
> 
> Having the UserAgent string of latest iPad OS would help.

Sure thing, thanks for looking into this.

Here's the user agent from an iPad 10th generation running iOS 17.5.1: `Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/17.5 Safari/605.1.15`

This UA results in new Device() throwing a "device not supported" exception.

ibc:
> Here's the user agent from an iPad 10th generation running iOS 17.5.1: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/17.5 Safari/605.1.15

Is that a **real and physical** iPad 20th gen device? Or are you running a simulator on computer?

nazar-pc:
User agent checking is an increasingly flawed approach, browsers reduce the amount of information in them and at some point you wouldn't be able to tell much at all. Maybe worth checking necessary APIs somehow instead for modern browsers?

ibc:
> Maybe worth checking necessary APIs somehow instead for modern browsers?

Even with that we should keep updating those checks for every new potential version of any device or browser. I basically regret that we do auto detection instead of forcing the parent application to do detection stuff (they already do in most of the apps) and choose a suitable mediasoup-client Handler programmatically (this is already possible of course, but many apps rely on mediasoup-client device auto detection anyway). I'm not sure I want to do same effort a different way.

nazar-pc:
I understand why it is convenient to use the library "as is" and have it "just work".

I'm wondering if it is possible to drop support for old versions of clients and if that allows to simplify checks to API exists/doesn't exist (I know a lot of stuff technically existed, but were broken in the past).

ibc:
I don't know which specific API checks you mean. Unfortunately that just covers a very few cases such as old Safari that didn't have RTCRtpTransceiver and things like that. Then Firefox supports simulcast but the order of encodings is the opposite to the order in other browsers. And tons of things like that that cannot be "detected via API checks" and instead we need to know if it's Firefox or not. Old Safari, Chrome, ReactNative and also EDGE (ORTC) handlers can be removed indeed, but that's not gonna solve any problem we have.

nazar-pc:
That is exactly what I'm talking about. Maybe time to forget old Safari. Also IIRC Firefox was meaning to fix the ordering, https://bugzilla.mozilla.org/show_bug.cgi?id=1663368 is closed for example.

So I'm wondering how many of those are still really relevant enough to support in the latest version of `mediasoup-client`?

ibc:
> That is exactly what I'm talking about. Maybe time to forget old Safari.

I insist that dropping old Safari (and other handlers) won't solve any problem here.

> Also IIRC Firefox was meaning to fix the ordering, https://bugzilla.mozilla.org/show_bug.cgi?id=1663368 is closed for example.

I guarantee that that bug was not fixed 5 years ago and last time I tried it in Firefox (a few months) ago it was still buggy.

iamarkdev:
> > Here's the user agent from an iPad 10th generation running iOS 17.5.1: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/17.5 Safari/605.1.15
> 
> Is that a **real and physical** iPad 20th gen device? Or are you running a simulator on computer?

This is my real, physical 10th generation iPad. Not a simulator.

ibc:
> This is my real, physical 10th generation iPad. Not a simulator.

Can you open this website in Safari on your iPad and paste here the detected UserAgent and the parsed data?

https://uaparser.dev/

iamarkdev:
> > This is my real, physical 10th generation iPad. Not a simulator.
> 
> Can you open this website in Safari on your iPad and paste here the detected UserAgent and the parsed data?
> 
> https://uaparser.dev/

Sure thing, here.

```json
{
  "ua": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/17.5 Safari/605.1.15",
  "browser": {
    "name": "Safari",
    "version": "17.5",
    "major": "17"
  },
  "cpu": {},
  "device": {
    "type": "tablet",
    "model": "iPad",
    "vendor": "Apple"
  },
  "engine": {
    "name": "WebKit",
    "version": "605.1.15"
  },
  "os": {
    "name": "macOS",
    "version": "10.15.7"
  }
}
```

Screenshot from my physical ipad: 

IMG_0106

ibc:
It's impossible that you get that result. I mean, if I paste your USerAgent "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/17.5 Safari/605.1.15" in that page I clearly get different results (the same result that mediasoup-client gets):

image

ibc:
Well, by looking at the code of ua-parser-js I see that it uses more data than just the UserAgent string. For example it uses [navigator.userAgentData](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/userAgentData) (although it's not supported by Safari at all) and more stuff such as `maxTouchPoints` and things like that.

Ok, I see that in order to enable those checks the library must be used a different way: https://docs.uaparser.dev/api/main/idata/with-feature-check.html

iamarkdev:
Thanks for looking into this and solving it @ibc 🙏

ibc:
working on it here https://github.com/versatica/mediasoup-client/pull/324

iamarkdev:
You guys are awesome 👍

ibc:
@iamarkdev please open https://v3demo.mediasoup.org in Safari in your iPad and let me know if it works and which mediasoup-client handler it detects (in the top left corner of the web):

example using Chrome in macbook:
image

iamarkdev:
Lgtm! 👍 
image

ibc:
mediasoup-client 3.10.0 released with the new APIs and documentation updated in the website. See the announcement in the mediasoup public forum:

https://mediasoup.discourse.group/t/mediasoup-client-3-10-0-released-with-new-and-better-browser-detection/6663

PR #282, opened 2023-10-26
TMP: for testing bug in CI Node 20

ibc:
https://github.com/versatica/mediasoup-client/actions/runs/6654886963/job/18084621753


nazar-pc:
You can still test this without a PR

PR #283, opened 2023-10-26
npm-scripts.ts: Fix replaceVersion() in Node 20

ibc:
Just that. Wrong usage of `fs.readdirSync()`.

### DIFF
--- a/npm-scripts.mjs
+++ b/npm-scripts.mjs
@@ -128,7 +128,8 @@ function replaceVersion()
 			continue;
 		}
 
-		const filePath = path.join('lib', file.name);
+		// NOTE: dirent.path is only available in Node >= 20.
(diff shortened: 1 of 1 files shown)


nazar-pc:
Hm, can about making it IIFE and read version from `package.json`? That seems to be a compatible change and doesn't require file edits.

ibc:
You mean using `fs.readfileSync(path.join(__dirname__, '..', 'package.json') and parse it etc? It's an overkill IMHO.

ibc:
BTW no no, this is mediasoup-client! Did you mean https://github.com/versatica/mediasoup/pull/1197?
In mediasoup-client we cannot read files! this may probably run in browser.

nazar-pc:
Right, I think I commented on the wrong PR :slightly_smiling_face:

PR #285, opened 2023-10-31
RtpParameters: enrich RtpParameters types

jmillan:
Since `flatbuffers` branch adds these new types in server side, we need to keep them in sync in case the application signalling relies in typed messages between client and server.


ibc (approved):

ibc:
I'll merge and release at the same time we release mediasoup.

PR #297, opened 2024-06-06
chore: add missing shadow dependency

ryuujo1573:
`npm` has a known problem of shadow dependencies. 
When installing with `pnpm`, the package `@types/node` is not there where `tsc` expects it, which itself is indirect dependency of `jest`.

And also, I would suggest use `pnpm` as a drop-in replacement for `npm`, which helps much on multi-package management, prevention of shadow deps, and quicker installation.

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,6 +78,7 @@
 	"devDependencies": {
 		"@types/events": "^3.0.3",
 		"@types/jest": "^29.5.12",
+		"@types/node": "20",
 		"@types/sdp-transform": "^2.4.9",
 		"@types/ua-parser-js": "^0.7.39",
 		"@typescript-eslint/eslint-plugin": "^7.10.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -23,6 +23,7 @@
 			"devDependencies": {
 				"@types/events": "^3.0.3",
 				"@types/jest": "^29.5.12",
+				"@types/node": "20",
 				"@types/sdp-transform": "^2.4.9",
 				"@types/ua-parser-js": "^0.7.39",
 				"@typescript-eslint/eslint-plugin": "^7.10.0",
@@ -1308,9 +1309,13 @@
 			"license": "MIT"
 		},
 		"node_modules/@types/node": {
-			"version": "18.11.9",
+			"version": "20.14.2",
+			"resolved": "https://registry.npmmirror.com/@types/node/-/node-20.14.2.tgz",
+			"integrity": "sha512-xyu6WAMVwv6AKFLB+e/7ySZVr/0zLCzOa7rSpq6jNwpqOrUbcACDWC+53d4n2QHOnDou0fbIsg8wZu/sxrnI4Q==",
 			"dev": true,
-			"license": "MIT"
+			"dependencies": {
+				"undici-types": "~5.26.4"
+			}
 		},
 		"node_modules/@types/sdp-transform": {
 			"version": "2.4.9",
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -78,6 +78,7 @@
 	"devDependencies": {
 		"@types/events": "^3.0.3",
 		"@types/jest": "^29.5.12",
+		"@types/node": "20",
```
ibc, line 81: I don't understand. Why do we need types/node dep and why do we need it in devDependencies? Which issue does this exactly fix?

ryuujo1573: Simply put, `jest` depends on `@types/node` type declarations, and it is standard implicit dependencies should be added. (Hoisted `node_modules` by `pnpm` default)

Running `tsc --noEmit` results in errors as follows: 

<details>
<summary>Logs</summary>

```
src/EnhancedEventEmitter.ts:112:2 - error TS2416: Property 'listeners' in type 'EnhancedEventEmitter<E>' is not assignable to the same property in base type 'EventEmitter'.
  Type '<K extends keyof E & string>(eventName: K) => Function[]' is not assignable to type '(type: string | number) => Listener[]'.
    Type 'Function[]' is not assignable to type 'Listener[]'.
      Type 'Function' is not assignable to type 'Listener'.
        Type 'Function' provides no match for the signature '(...args: any[]): void'.

112  listeners<K extends keyof E & string>(eventName: K): Function[] {
     ~~~~~~~~~

src/EnhancedEventEmitter.ts:116:2 - error TS2416: Property 'rawListeners' in type 'EnhancedEventEmitter<E>' is not assignable to the same property in base type 'EventEmitter'.
  Type '<K extends keyof E & string>(eventName: K) => Function[]' is not assignable to type '(type: string | number) => Listener[]'.
    Type 'Function[]' is not assignable to type 'Listener[]'.

116  rawListeners<K extends keyof E & string>(eventName: K): Function[] {
     ~~~~~~~~~~~~

src/test/test.ts:1723:28 - error TS2304: Cannot find name 'global'.

1723  const originalNavigator = global.navigator;
                                ~~~~~~

src/test/test.ts:1734:39 - error TS2304: Cannot find name 'global'.

1734     const originalRTCRtpTransceiver = global.RTCRtpTransceiver;
                                           ~~~~~~

src/test/test.ts:1737:6 - error TS2304: Cannot find name 'global'.

1737      global.RTCRtpTransceiver = class Dummy {
          ~~~~~~

src/test/test.ts:1745:5 - error TS2304: Cannot find name 'global'.

1745     global.RTCRtpTransceiver = originalRTCRtpTransceiver;
         ~~~~~~

src/test/test.ts:1752:2 - error TS2304: Cannot find name 'global'.

1752  global.navigator = originalNavigator;
      ~~~~~~


Found 7 errors in 2 files.

Errors  Files
     2  src/EnhancedEventEmitter.ts:112
     5  src/test/test.ts:1723
```

</details>

In another sense, since `global` is used, then type definition for `global` is required, and runtime bindings are provided by `node`. I reckon it as one of `devDependencies`.

ibc:
Thanks. I will merge soon but won't be able to release until next week.

ibc:
CI is failing. You must run npm install types/node and commit

ryuujo1573:
> CI is failing. You must run npm install types/node and commit

Sorry for missing this.

I've added the corresponding commit to the branch. Maybe we can squeeze it for clarity 😅

ibc:
I need to wait a bit before merging this because we have a related issue that may need additional changes. Will be done soon.

PR #300, opened 2024-06-12
Allow createEncodedStreams on PCs without encodedInsertableStreams parameter

### LINKED ISSUE #299: Allow createEncodedStreams on PCs without encodedInsertableStreams parameter
### Background

Mediasoup supports an additionalSetting on transports encodedInsertableStreams: true which is a Chromium-specific extension allowing for an equivalent to [RTCRTPScriptTransform](https://developer.mozilla.org/en-US/docs/Web/API/RTCRtpScriptTransform) to be added to senders and receivers.

When encodedInsertableStreams is true, applications can call createEncodedStreams() on RTCRtpReceiver/RTCRtpSender at any time after creation. Until this function is called, all packets in the associated streams are dropped. Thus, in order to flow data, a [RTCRTPScriptTransform](https://developer.mozilla.org/en-US/docs/Web/API/RTCRtpScriptTransform) must be registered for each stream coming from the peerconnection. This is unfortunate, because there is performance overhead from these streams and there is no way to not-set a transform (for example, if we don't want to transform video streams).

Here the W3 spec: https://www.w3.org/TR/webrtc-encoded-transform/

### What's New?

Starting with Chrome 121 (~Feb 2024), the Google team landed a change [Allow createEncodedStreams on PCs without encodedInsertableStreams param](https://chromium-review.googlesource.com/c/chromium/src/+/5077034) which, as the title says, allows creating a [RTCRTPScriptTransform](https://developer.mozilla.org/en-US/docs/Web/API/RTCRtpScriptTransform) on streams created from PCs that don't set the encodedInsertableStreams flag. This is perfect, since we can register transforms only for the actual streams which need it.

The limitation that comes along with the new capability is that if a transform is needed, createInsertableStreams() must be called synchronously "[within one JS event loop spin [of creating the receiver/sender]](https://issues.chromium.org/issues/40943169#comment7)".

### Problem

mediasoup-client abstracts the process of sender/receiver creation behind async Transport.producer/receive() functions. Thus, by the time these functions return it is too late to register the transform. Indeed Chrome gives the following error: InvalidStateError: Failed to execute 'createEncodedStreams' on 'RTCRtpReceiver': Too late to create encoded streams.

The proper place where createEncodedStreams() must be called (for a receiver) is here: https://github.com/versatica/mediasoup-client/blob/470c90d4cd61d203c8345e931702aee65d6308e7/src/handlers/Chrome111.ts#L787, just after the line:

```ts
// src/handlers/Chrome111.ts @ receive()

await this._pc.setRemoteDescription(offer);

// NEW: Create the transform streams.
for (const options of optionsList) {
    const { trackId } = options;
    const localId = mapLocalId.get(trackId);
    const transceiver = this._pc.getTransceivers()
        .find((t) => t.mid === localId);
    if (!transceiver) {
        throw new Error('transceiver not found');
    }
    const {
        readable,
        writable,
    } = transceiver.receiver.createEncodedStreams();
}
```

But there is the problem of how to hand the readable, writable streams back out to the caller. Additionally, this is a Chromium-specific API and therefore these streams will never exist on Firefox nor Safari.

One alternative is to hook from the application into the track event of the internal PC before calling receive(), but this goes against existing mediasoup API design:

``` ts
// Application code, during transport creation.

newTransport.handler._pc.addEventListener(
  'track',
  (event) => {
    const {
      readable,
      writable,
    } = event.transceiver.receiver.createEncodedStreams();
  },
);
```

### Proposed Solution

We could add a new callback to [ConsumerOptions](https://mediasoup.org/documentation/v3/mediasoup-client/api/#ConsumerOptions) so the application can pass an optional onRtpReceiver callback:

``` ts
const consumer = await recvTransport.consume({
  xxxxx,
  xxxxx,
  onRtpReceiver: (receiver) => {
    const { readable, writable } = receiver.createEncodedStreams();
    // do things with them
  }
});
```

In Chrome handler:

``` ts
await this._pc.setRemoteDescription(offer);

for (const options of optionsList) {
  const { trackId, onRtpReceiver } = options;

  if (onRtpReceiver) {
    const localId = mapLocalId.get(trackId);
    const transceiver = this._pc.getTransceivers()
      .find((t) => t.mid === localId);
    
    if (!transceiver) {
      throw new Error('transceiver not found');
    }

    onRtpReceiver(transceiver.receiver);
  }
}
```

Also, add similar logic to the `sendTransport.produce` function for registering sender side equivalents.

kjvenalainen:
Fixes #299 

Allows calling clients to add sender and receiver callback functions which are executed synchronously after `RTCRTPSender/Receiver` creation. This feature is targeted to Chromium browsers and allows applications to call `createEncodedStreams()` in the callback functions in order to create [WebRTC Encoded Transforms](https://www.w3.org/TR/webrtc-encoded-transform/).

### Example Usage

``` ts
const consumer = await recvTransport.consume({
  xxxxx,
  xxxxx,
  onRtpReceiver: (receiver) => {
    const { readable, writable } = receiver.createEncodedStreams();
    readable.pipeTo(writable);
  }
});
```

### DIFF
--- a/src/handlers/HandlerInterface.ts
+++ b/src/handlers/HandlerInterface.ts
@@ -34,11 +34,18 @@ export type HandlerRunOptions = {
 	extendedRtpCapabilities: any;
 };
 
+/*
+ * Invoked synchronously immediately after a new RTCRtpSender is created.
+ * This allows for creating encoded streams in chromium browsers.
+ */
+export type SenderCallback = (sender: RTCRtpSender) => void;
+
 export type HandlerSendOptions = {
 	track: MediaStreamTrack;
 	encodings?: RtpEncodingParameters[];
 	codecOptions?: ProducerCodecOptions;
 	codec?: RtpCodecCapability;
+	onRtpSender?: SenderCallback;
 };
 
 export type HandlerSendResult = {
@@ -46,6 +53,12 @@ export type HandlerSendResult = {
 	rtpParameters: RtpParameters;
 	rtpSender?: RTCRtpSender;
 };
+
+/*
+ * Invoked synchronously immediately after a new RTCRtpReceiver is created.
+ * This allows for creating encoded streams in chromium browsers.
+ */
+export type ReceiverCallback = (receiver: RTCRtpReceiver) => void;
 
 export type HandlerReceiveOptions = {
 	trackId: string;
@@ -58,6 +71,7 @@ export type HandlerReceiveOptions = {
 	 * can just synchronize up to one audio stream with one video stream.
 	 */
 	streamId?: string;
+	onRtpReceiver?: ReceiverCallback;
 };
 
 export type HandlerReceiveResult = {
--- a/src/Consumer.ts
+++ b/src/Consumer.ts
@@ -2,7 +2,7 @@ import { Logger } from './Logger';
 import { EnhancedEventEmitter } from './EnhancedEventEmitter';
 import { InvalidStateError } from './errors';
 import { MediaKind, RtpParameters } from './RtpParameters';
-import { AppData } from './types';
+import { AppData, ReceiverCallback } from './types';
 
 const logger = new Logger('Consumer');
 
@@ -12,6 +12,7 @@ export type ConsumerOptions<ConsumerAppData extends AppData = AppData> = {
 	kind?: 'audio' | 'video';
 	rtpParameters: RtpParameters;
 	streamId?: string;
+	onRtpReceiver?: ReceiverCallback;
 	appData?: ConsumerAppData;
 };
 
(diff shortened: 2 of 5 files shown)


src/handlers/HandlerInterface.ts
```diff
@@ -34,11 +34,18 @@ export type HandlerRunOptions = {
 	extendedRtpCapabilities: any;
 };
 
+/*
```
ibc, line 37: ** please

src/handlers/HandlerInterface.ts
```diff
@@ -47,6 +54,12 @@ export type HandlerSendResult = {
 	rtpSender?: RTCRtpSender;
 };
 
+/*
```
ibc, line 57: Same

jmillan (approved):

ibc:
I tried to push to this repo following GH instructions and... ended creating another PR #301 by accident (not sure what I did wrong). Let's continue in there. Sorry.

PR #308, opened 2024-10-10
Export TS types and extras in a separate module entry point

ibc:
## Details

- Instead of exporting `types` like this:
  ```ts
  import { types as mediasoupClientTypes } from 'mediasoup-client';
  ```
- Now we import them this way:
  ```ts
  import * as mediasoupClientTypes from 'mediasoup-client/types';
  ```
- Also export new `extras` namespaces in both ways:
  ```ts
  import { extras } from 'mediasoup-client';
  ```
- And this way:
  ```ts
  import { HandlerInterface, HandlerFactory } from 'mediasoup-client/extras/HandlerInterface';
  import { FakeHandler } from 'mediasoup-client/extras/FakeHandler';
  import * as mediasoupClientTestFakeParameters from 'mediasoup-client/extras/fakeParameters';
  ```

## NOTE 1

It looks like for this to work, the parent application must have in its `tsconfig.json`:

```
"module": "nodenext",
"moduleResolution": "nodenext",
```

But take into account that using "moduleResolution: node" means "Node 10". Please, we can kill that already.

## NOTE 2

Should we stop exporting `types` from 'mediasoup-client' and instead force applications to import them from 'mediasoup-client/types'?

Well, we don't do it yet.

### DIFF
--- a/.github/workflows/mediasoup-client.yaml
+++ b/.github/workflows/mediasoup-client.yaml
@@ -19,9 +19,11 @@ jobs:
             node: 20
           - os: ubuntu-24.04
             node: 22
-          - os: macos-12
-            node: 20
+          - os: macos-13
+            node: 18
           - os: macos-14
+            node: 20
+          - os: macos-15
             node: 22
           - os: windows-2022
             node: 22
--- a/package-lock.json
+++ b/package-lock.json
@@ -27,8 +27,8 @@
 				"@types/node": "22",
 				"@types/sdp-transform": "^2.4.9",
 				"@types/ua-parser-js": "^0.7.39",
-				"@typescript-eslint/eslint-plugin": "^8.18.0",
-				"@typescript-eslint/parser": "^8.18.0",
+				"@typescript-eslint/eslint-plugin": "^8.18.1",
+				"@typescript-eslint/parser": "^8.18.1",
 				"eslint": "^9.17.0",
 				"eslint-config-prettier": "^9.1.0",
 				"eslint-plugin-jest": "^28.9.0",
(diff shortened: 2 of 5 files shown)


jmillan:
> I've briefly tested this and it looks like for this to work, the parent application must have in its tsconfig.json:

>"module": "nodenext",
>"moduleResolution": "nodenext",

If the application does not set such values (for any reason they may have), can it still use the old way of importing the types?

ibc:
I am not sure. I've tried to test using mediasoup-demo client app but it's SUPER OLD and doesn't even use TypeScript so we are lost here.

ibc:
This is ready for review.

jmillan (approved):

ibc:
**NOTE:** These changes have been reverted in 3.8.1. The world is not ready for these things yet.

Fabioni:
> **NOTE:** These changes have been reverted in 3.8.1. The world is not ready for these things yet.

What's the reason?

ibc:
> > **NOTE:** These changes have been reverted in 3.8.1. The world is not ready for these things yet.
> 
> 
> 
> What's the reason? 

The reason is that the syntax of the "exports" field in package.json is not universally defined. NPM docs refer to the documentation of "exports" in package.json of Node projects where the value of each item in "exports" is basically a path to the corresponding file (it doesn't cover TypeScript .d.ts files at all), while in webpack the value can be an object with "import", "require", "types" (pointing to the .d.ts file with TypeScript definitions). Then ts-jest doesn't read the "types" field so it throws an error because "types not found", etc. Then ts-jest also fails if an entry in "exports" points to a file that exports a const, etc etc. I investigated and indeed it's a know ts-jest issue.

It's a complete nightmare and I already spent too much time trying to deal with it.

PR #311, opened 2024-11-10
Add support for sending abs-capture-time RTP extension

jbaudanza:
I updated this draft PR with support for react-native https://github.com/versatica/mediasoup-client/pull/217

The reason I need this feature is I'm building a voice chat-bot that using using webrtc/mediasoup and has a "start/stop" record button. For this, I need precise timestamps to capture exactly what the user has said.

We might want to add this to other handlers as well, like Chrome111.


ibc:
I merged a similar PR yesterday and released a new version. Thanks.

PR #312, opened 2024-11-26
Fix sprop-stereo handling

ggarber:
When sprop-stereo is "0" we are still assuming that is enabled and enabling stereo decoding.

Note: This fix could break stereo working for some implementations.


ibc:
Wow. Checking this morning and will release later.

ibc:
> Note: This fix could break stereo working for some implementations.

It could just break scenarios where some NON mediasoup-client sender is sending stereo audio without properly signaling it in `RtpParameters.codecs['opus']`, right?

ibc:
Released in mediasoup-client 3.7.18. Thanks

ggarber:
> > Note: This fix could break stereo working for some implementations.
> 
> It could just break scenarios where some NON mediasoup-client sender is sending stereo audio without properly signaling it in `RtpParameters.codecs['opus']`, right?

If the server was signaling "sprop-stereo=0" before stereo was still working.   Now with those servers it will stop working.

ibc:
> If the server was signaling "sprop-stereo=0" before stereo was still working. Now with those servers it will stop working.

The server will send "sprop-stereo=0" or "=1" based on client given `RtpParameters` so we are good.

PR #320, opened 2025-03-26
Add AV1 codec support

jmillan:
- Chrome111: Retrieve Dependency Descriptor header extension.
- Firefox120: Remove `scalabilityMode` since it does not support SVC.
- RemoteSdp: Remove Dependency Descriptor header extension for any code but AV1 and H264.


src/handlers/sdp/RemoteSdp.ts
```diff
@@ -172,6 +174,24 @@ export class RemoteSdp {
...
+		);
+
+		if (!ddCodec) {
+			// @ts-expect-error --- mediaObject type is 'any'.
+			mediaObject.ext = mediaObject.ext.filter(
+				(extmap: any) =>
+					extmap.uri !==
+					'https://aomediacodec.github.io/av1-rtp-spec/#dependency-descriptor-rtp-header-extension'
+			);
+		}
```
ibc, line 193: My initial expectation was that this should be done by calling some útil function from each handler but now I understand that this is not related to each handler but to mediasoup server which only supports DD for AV1. 

So despite I don't like that this is hardcoded in client I understand that we don't signal anything like "only AV1 supports DD" from mediasoup to client since that info can not be expressed via Router capabilities. So I think this is indeed the best we can do for now.

ibc (approved):

src/handlers/sdp/RemoteSdp.ts
jmillan: Yes, basically we are here adapting the capabilities to what mediasoup expects. Yes, we support DD extension but only for AV1 for now.

PR #322, opened 2025-04-08
Allow receiving 1 and 2 bytes header extensions

jmillan:
Since mediasoup will send both types.

Enable at both session and media level.


src/handlers/sdp/RemoteSdp.ts
```diff
@@ -206,6 +206,8 @@ export class RemoteSdp {
 			mediaSection = this._mediaSections[idx] as OfferMediaSection;
 		}
```
ibc, line 208: Comment please?

ibc (approved):

PR #331, opened 2025-07-23
Remove old handlers

ibc:
## Details

- Remove old handlers (those more than 6 years old).
- This includes `ReactNative` (Plan-B), so rename `ReactNativeUnifiedPlan` to `ReactNative106` (breaking change, I know, but all this PR is a breaking change). The thing is that we cannot maintain forever those super old files.
- As a result, remove Plan-B support entirely.

## TODO

- [x] Update `BuiltinHandlerName` TS type in the API documentation in the website.
- [x] Announce it in the website (breaking changes).


jmillan (approved):
Relief... 👌👌

PR #334, opened 2025-08-12
fix: peerconnection memory leak

zt123123:


src/handlers/Chrome111.ts
```diff
@@ -251,6 +251,8 @@ export class Chrome111 extends HandlerInterface {
 		if (this._pc) {
 			try {
 				this._pc.close();
+				// @ts-ignore
+				this._pc = null;
```
ibc, line 255: Why this? Why do we need to set the peerconnection to `null`? How does it prevent any memory leak? This PR has literally zero text description or rationale. 

You may have a memory leak in your application but that could be related to the fact that you create many mediasoup-client `Transports` and never un-reference them, so they leak, and hence the peerconnection they contain also leaks. But that's not a bug in mediasoup-client.

Also, why `@ts-ignore`? Have you taken any look at the code? We don't use `@ts-ignore` anywhere.

ibc: And of course the `lint` task fails due to the usage of `@ts-ignore`: https://github.com/versatica/mediasoup-client/actions/runs/16900979808/job/47890263603?pr=334

Anyway, this is not the point. Why and how is this PR fixing a memory leak?

zt123123: the same issue in my project, look this [https://stackoverflow.com/questions/76474381/domexception-failed-to-construct-rtcpeerconnection-cannot-create-so-many-pee](cannot-create-so-many-pee)

zt123123: I referred to this post [I referred to this post https://stackoverflow.com/questions/49532939/how-to-clear-rtcpeerconnection-webrtc/49644693](https://stackoverflow.com/questions/49532939/how-to-clear-rtcpeerconnection-webrtc/49644693)

ibc: It's not true that pc must be set to `null` in order to be garbage collected. Once `pc.close()` is called, GC will erase `pc` if `pc` is not referenced anywhere.

ibc:
@zt123123, I'm replacing this PR with this one: https://github.com/versatica/mediasoup-client/pull/335

zt123123:
> @zt123123, I'm replacing this PR with this one: #335

okay tks

PR #336, opened 2025-08-13
Don't expect static codec payload types and RTP extension IDs

### LINKED ISSUE #330: Produce Video only stream fails.
### Your environment

- Operating system: Mac OS Sequoia 15.5
- Browser version: Chrome Dev - Version 140.0.7259.2 (Official Build) dev (x86_64)
- npm version: 10.8.2
- mediasoup version:  "github:versatica/mediasoup#v3" - [13d05b4]
- mediasoup-client version: "github:versatica/mediasoup-client#v3" - [0e89d6b]

### Issue description
Produce Video only stream fails with the error 

`failed:InvalidAccessError: Failed to execute 'setRemoteDescription' on 'RTCPeerConnection': Failed to set remote answer sdp: Failed to set recv parameters for m-section with mid='0'
`

Steps to recreate:
Able to recreate this in media soup-demo by not enabling mic when joining the room. Have attached the git diff
[video_only.diff.txt](https://github.com/user-attachments/files/21342454/video_only.diff.txt)

_Note: This issue happens only in the beta and dev versions of Chrome, but works fine in Chrome stable Version 138.0.7204.158 (Official Build) (x86_64)_

ibc:
## Details

- Fixes #330
- In Chrome M140 the way codec payload types and RTP extension IDs are assigned in the SDP will change and will break assumptions we do in mediasoup-client.
- Before this PR, mediasoup-client obtains local RTP capabilities when `device.load()` is called and assumes that the codec payload types and RTP extension IDs will be later the same ones when sending `Transports` are created and `Producers` are created on them. So for example, in Chrome M140, if we start with video and then we add audio to a sending `Transport`, those numeric payload types and extension local IDs will be different than the ones in our discovered RTP capabilities, and hence our sending `RtpParameters` will have wrong values that won't match what the browser will really send into the RTP packets, so everything will fail.
- In this PR, mediasoup-client computes local capabilities every time `transport.produce()` is called and obtains the codec payload types and RTP extension IDs dynamically from the latest SDP offer generated by the browser.
- Thanks a lot to @ramesh-kv-rc for properly reporting the issue.

## Important

- **Upgrading to next mediasoup-client version will be urgently needed, otherwise many apps will stop working with errors like this:**
  ```
  failed:InvalidAccessError: Failed to execute 'setRemoteDescription' on 'RTCPeerConnection': Failed to set remote answer sdp: Failed to set recv parameters for m-section with mid='0'
  ```
- The new dynamic RTP extension ID assign logic in libwebrtc has already landed in latest Chrome Canary 141.0.7354.0 and indeed mediasoup-client `v3` branch **fails** on it. However by using the branch in this PR it works fine.

### DIFF
--- a/src/Device.ts
+++ b/src/Device.ts
@@ -91,13 +91,18 @@ export class Device {
 	private readonly _handlerName: string;
 	// Loaded flag.
 	private _loaded = false;
-	// Extended RTP capabilities.
-	private _extendedRtpCapabilities?: ExtendedRtpCapabilities;
+	// Callback for sending Transports to request sending extended RTP capabilities
+	// on demand.
+	private _getSendExtendedRtpCapabilities?: (
+		nativeRtpCapabilities: RtpCapabilities
+	) => ExtendedRtpCapabilities;
 	// Local RTP capabilities for receiving media.
 	private _recvRtpCapabilities?: RtpCapabilities;
-	// Whether we can produce audio/video based on computed extended RTP
-	// capabilities.
-	private readonly _canProduceByKind: CanProduceByKind;
+	// Whether we can produce audio/video based on remote RTP capabilities.
+	private readonly _canProduceByKind: CanProduceByKind = {
+		audio: false,
+		video: false,
+	};
 	// Local SCTP capabilities.
 	private _sctpCapabilities?: SctpCapabilities;
 	// Observer instance.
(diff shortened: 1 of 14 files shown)


ibc:
New version 3.15.0 released and announced in the forum:

https://mediasoup.discourse.group/t/urgent-upgrade-to-mediasoup-client-3-15-0-asap/6821

PR #349, opened 2025-09-24
Nullish Coalescing Assignment operator replacement

msavin-pyrus:
Operator _Nullish Coalescing Assignment_ (??=) is not supported in older browsers below Chrome 87 and causes a syntax error `Uncaught SyntaxError: Unexpected token '='`.

PR replaces operator call with two separate operators:
`variable ??= expression` => `variable = variable ?? expression`


ibc:
Since we are supposed to support Chrome 74, this PR makes sense. Thanks.

ibc:
Released in mediasoup-client 3.16.6.

PR #355, opened 2025-12-10
Fix regression in Chrome 143 due to changes in RTP extensions handling

### LINKED ISSUE #353: VP8 simulcast Producer fails on second creation (setLocalDescription error) after webcam toggle (Chrome 143)
# Original issue description

I’m running into a reproducible VP8 simulcast issue in Chrome when a user turns their webcam on/off. I’d like to understand whether this is a known Chrome/libwebrtc issue, a mediasoup-client behavior, or something in my flow.

Environment

- mediasoup: 3.19.13
- mediasoup-client: 3.18.1

Browsers tested:

- Chrome 143 → problem occurs
- Firefox 145 → works correctly

Scenario:

- Video codec: VP8 with simulcast enabled
- One WebRtcTransport for sending
- No SDP munging

When a user:
1. Enters the room with webcam ON, creating the first VP8 simulcast producer
2. Turns webcam OFF (I close the producer)
3. Turns webcam ON again (new VP8 simulcast producer on the same send transport)
4. Chrome fails during its internal negotiation with:
  ```
  InvalidAccessError: Failed to execute ‘setLocalDescription’ on ‘RTCPeerConnection’:
  Failed to set local offer sdp: The media section with MID=‘0’ negotiates simulcast
  but does not negotiate the RID RTP header extension.
  ```

This error happens before any interaction with the mediasoup server — Chrome rejects its own offer inside mediasoup-client.

Important additional observation

The same flow works perfectly in Firefox 145.

Other users are also affected: [https://mediasoup.discourse.group/t/vp8-simulcast-producer-fails-on-second-creation-setlocaldescription-error-after-webcam-toggle-chrome-143/6925](https://mediasoup.discourse.group/t/vp8-simulcast-producer-fails-on-second-creation-setlocaldescription-error-after-webcam-toggle-chrome-143/6925)

# STR in the mediasoup-demo

1. Join a room without mic and without datachannel so video will become the first `m=` section: `https://local.dev:5555/?roomId=xxx&info=true&stats=false&consume=false&datachannel=false&mic=false`
2. Have camera enabled.
3. Disable camera.
4. Enable camera again.
5. `enableWebcam()` method in the demo app fails with "InvalidAccessError: Failed to execute 'setLocalDescription' on 'RTCPeerConnection': Failed to set local offer sdp: The media section with MID='0' negotiates simulcast but does not negotiate the RID RTP header extension".

# More info

This happens due to a change in libwebrtc in the way it handles RTP extensions in the SDP. It only affects to the first `m=` section in the SDP because we **never** close it (if we closed it then the ICE transport would disconnect due to WebRTC Bundle mess).

Changes in libwebrtc that trigger this problem:
- https://developer.chrome.com/release-notes/143#webrtc_rtp_header_extension_behavior_change
- https://issues.webrtc.org/issues/439514253

Tickets in Chromium about this same problem (affects to many people):
- https://issues.chromium.org/issues/467164231
- https://issues.chromium.org/issues/467468926

ibc:
# Details

- Fixes #353
- Tested in Chrome 143.
- Tested in Opera 122 which uses Chromium version:138.0.7204.251.
- Tested in Firefox 144.0.2.
- Tested in Safari 18.6.

## Rationale

Chrome 143 complains when we disable the first m= section in the sending PeerConnection. It complains because its locally generated offer contains simulcast attributes but the SDP answer mediasoup-client generates doesn't contain the corresponding RTP extensions for simulcast (RID). And since we are not closing the transceiver (because it's the first one so it would disconnect ICE), we only "disable" it which means that the response has "a=inactive" but the lack of RID extension makes Chrome 143 throw an error because the offer requests simulcast. Solution is to NOT remove RTP extension attributes from the disabled m= section in the remote SDP answer, and only do it when closing the transceiver (which involves remote port set to 0).

### DIFF
--- a/src/handlers/sdp/MediaSection.ts
+++ b/src/handlers/sdp/MediaSection.ts
@@ -107,7 +107,12 @@ export abstract class MediaSection {
 
 	disable(): void {
 		this.pause();
+	}
 
+	close(): void {
+		this.disable();
+
+		this._mediaObject.port = 0;
 		delete this._mediaObject.ext;
 		delete this._mediaObject.ssrcs;
 		delete this._mediaObject.ssrcGroups;
@@ -115,12 +120,6 @@ export abstract class MediaSection {
 		delete this._mediaObject.simulcast_03;
 		delete this._mediaObject.rids;
 		delete this._mediaObject.extmapAllowMixed;
-	}
-
-	close(): void {
-		this.disable();
-
-		this._mediaObject.port = 0;
 	}
 }
 
--- a/src/handlers/sdp/MediaSection.ts
+++ b/src/handlers/sdp/MediaSection.ts
@@ -112,8 +112,19 @@ export abstract class MediaSection {
 	close(): void {
 		this.disable();
 
+		// Set port in m= line to 0, which means that the media sction is closed.
 		this._mediaObject.port = 0;
-		delete this._mediaObject.ext;
+
+		// NOTE: Do not remove header extensions since it's controversial in the spec.
+		delete this._mediaObject.candidates;
+		delete this._mediaObject.endOfCandidates;
+		delete this._mediaObject.iceUfrag;
+		delete this._mediaObject.icePwd;
+		delete this._mediaObject.iceOptions;
+		this._mediaObject.rtp = [];
+		this._mediaObject.fmtp = [];
+		delete this._mediaObject.rtcp;
+		delete this._mediaObject.rtcpFb;
 		delete this._mediaObject.ssrcs;
 		delete this._mediaObject.ssrcGroups;
 		delete this._mediaObject.simulcast;


ibc:
Notice that this is not a FULL solution. See my comment in the libwebrtc ticket: https://issues.chromium.org/issues/467164231#comment20

Astagor:
Thank you for your work. So to have a full solution, we need to wait for Chromium devs, right?

ibc:
> Thank you for your work. So to have a full solution, we need to wait for Chromium devs, right?

Yes. I've been told that they are gonna revert the change ASAP.

Astagor:
So I will wait and not change anything yet, as my patch (audio before video) works for me in most cases. And we never close audio producer but pause, we only close video producers.

jmillan (approved):

ibc:
mediasoup-client 3.18.2 released with the fix. Tested in many browsers.

PR #359, opened 2026-02-17
Handle sending and receiving RTP capabilities separately

### LINKED ISSUE #141: Incorrect reporting for the RTP capabilities if the 'Device' is used only for consuming streams
### Issue description

Some android devices support decoding h264 but does not support encoding it, 
Most Huawei devices are like this (even the modern ones)

These devices response only vp8/vp9 support for the following code 

```
const pc = new RTCPeerConnection(
			{
				iceServers         : [],
				iceTransportPolicy : 'all',
				bundlePolicy       : 'max-bundle',
				rtcpMuxPolicy      : 'require',
				sdpSemantics       : 'unified-plan'
			});
pc.addTransceiver('audio');
pc.addTransceiver('video');

const offer = await pc.createOffer();

console.log(offer.sdp)
```

which is what is used in the Chrome74 Device Handler 

however response with some h264 support (in addition to the vp8/vp9 support) using the following code

```
const pc = new RTCPeerConnection(
			{
				iceServers         : [],
				iceTransportPolicy : 'all',
				bundlePolicy       : 'max-bundle',
				rtcpMuxPolicy      : 'require',
				sdpSemantics       : 'unified-plan'
			});

const at = pc.addTransceiver('audio');
const vt = pc.addTransceiver('video');

// you can add the following to check the decoding capabilities
at.direction = "recvonly"
vt.direction = "recvonly"

const offer = await pc.createOffer();

console.log(offer.sdp)
```
notice that setting the direction to 'recvonly' cause thee sdp offer to response differently 

so in case you want to consume a h264 stream even though the h264 decoding is supported, the device response with empty list  for the RTP capabilities causing the mediasoup server router to return false for the canConsume 
```
router.canConsume({
            producerId,
            rtpCapabilities,
        })
```

resulting in inability for the user to consume the stream 

right now in our application we have a patch for mediasoup-client to accept a direction which can be `'sendonly' | 'recvonly' | 'sendrecv'` and return the RTP capabilities accordingly resulting in a functional consumtion 

here is a short list of some of the devices which we found have this issue 
```
Samsung SM-G532F
Samsung SM-J320H
CASPER_VIA_M3
Etab5
HRY-LX1T
HTC Desire 830
Huawei ANE-LX1
Huawei ANE-LX2
Huawei DRA-LX2
Huawei ELE-L29
Huawei FIG-LX1
Huawei JSN-L21
Huawei MRD-LX1
Huawei POT-LX1
Huawei SNE-LX1
Huawei VNS-L31
Huawei VOG-L29
INE-LX2r
LG-D693TR
LG-K430
M2004J19C
MAR-LX1A
```

we tested the patch and it worked for what we have tested up to now though we still haven't finish testing all of them

6uliver:
@ibc as you mentioned in this ticket (#141):

"The device can later create send transports and recv transports. There is where we should run different code to get capabilities from a local SDP offer. But unfortunately we don't do that (yet)."

As the library has separate transports we don't need to set the direction in the Device constructor anymore but we can have a direction parameter in the getNativeRtpCapabilities. This will solve the problem mentioned in #141 that our team has as well: if the device can not encode (send) something the library won't detect the capability for decoding (receiving). By setting the direction for the transceivers in the getNativeRtpCapabilities we will get the correct capabilities for sending and receiving which are possibly different.

I tried to keep the public interface backward compatible (rtpCapabilities is doing the same as before but with the fix so it's really the capabilities for receiving media) and extend it with new functionalities: rtpReceiveCapabilities, rtpSendCapabilities.
The getRecvRtpCapabilities is working as before, I've just refactored it to a helper function and added the new getSendRtpCapabilities.

I hope you like it and can merge as soon as possible. In case you have any requests feel free to ask, I'm more than happy to fix anything related to this problem.

### DIFF
--- a/src/Device.ts
+++ b/src/Device.ts
@@ -134,6 +134,8 @@ export class Device {
 	) => ExtendedRtpCapabilities;
 	// Local RTP capabilities for receiving media.
 	private _recvRtpCapabilities?: RtpCapabilities;
+	// Local RTP capabilities for sending media.
+	private _sendRtpCapabilities?: RtpCapabilities;
 	// Whether we can produce audio/video based on remote RTP capabilities.
 	private readonly _canProduceByKind: CanProduceByKind = {
 		audio: false,
@@ -260,14 +262,38 @@ export class Device {
 	/**
 	 * RTP capabilities of the Device for receiving media.
 	 *
+	 * @deprecated Use {@link rtpReceiveCapabilities} instead.
+	 *
 	 * @throws {InvalidStateError} if not loaded.
 	 */
 	get rtpCapabilities(): RtpCapabilities {
+		return this.rtpReceiveCapabilities;
+	}
+
+	/**
+	 * RTP capabilities of the Device for receiving media.
+	 *
+	 * @throws {InvalidStateError} if not loaded.
+	 */
+	get rtpReceiveCapabilities(): RtpCapabilities {
 		if (!this._loaded) {
 			throw new InvalidStateError('not loaded');
 		}
 
 		return this._recvRtpCapabilities!;
+	}
+
+	/**
+	 * RTP capabilities of the Device for sending media.
+	 *
+	 * @throws {InvalidStateError} if not loaded.
+	 */
+	get rtpSendCapabilities(): RtpCapabilities {
+		if (!this._loaded) {
+			throw new InvalidStateError('not loaded');
+		}
+
+		return this._sendRtpCapabilities!;
 	}
 
 	/**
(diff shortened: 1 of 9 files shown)


ibc:
Thanks @6uliver. It looks good but honestly this change is too sensitive and I need to redo it on my own to be sure of what these changes do. Probably I'll end copy&pasting 99% of your PR but I need to do it. Let me please check these changes and work on this these days.

ibc:
Hi @6uliver, as said above I basically forked your PR, did minor cosmetic changes and I tested it locally with all browsers. Let me replace this PR.

https://github.com/versatica/mediasoup-client/pull/360

6uliver:
Great, thanks for your quick response!

PR #362, opened 2026-04-01
fix: s.brands.find is not a function error

evgmel:
**Details**

Some browsers throw `(s.brands ?? []).find is not a function` error when a new Device is created. 

That error prevents people with such browsers from starting any webrtc call.

The fix changes the way we check a browser brands, checking exactly that `brands` property is array, not only null/undefined.

### DIFF
--- a/src/Device.ts
+++ b/src/Device.ts
@@ -706,7 +706,10 @@ function getChromiumMajorVersion(
 		// Some nasty browser extensions define their own custom
 		// navigator.userAgentData without mandatory `brands` field, so let's be
 		// ready for it.
-		const chromiumBrand = (userAgentData.brands ?? []).find(
+		const brands = Array.isArray(userAgentData.brands)
+			? userAgentData.brands
+			: [];
+		const chromiumBrand = brands.find(
 			b => b.brand === 'Chromium'
 		);
 


nazar-pc (approved):

evgmel:
Hi @ibc ! Could you please check the PR with the fix?

ibc:
>  Hi @ibc ! Could you please check the PR with the fix? 

Not really. I am on vacation.

jmillan:
@evgmel, just to get more context, can you provide the following info:

- What device is suffering this issue.
- What exactly is: `userAgentData.brands` on such device?

evgmel:
@jmillan Unfortunately, we do not gather the detailed values for `brands` on the devices, except user agents. According to the UA values it's just ordinary web browsers:

- Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/146.0.0.0 Safari/537.36
- Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/146.0.0.0 Safari/537.36 Edg/146.0.0.0

Our guess is, may be some some browser extension can impact `brands` value to be of different format.
Most probably it's `string` (not undefined of null, because in this case the error text would be different).

ibc:
Thanks. Released in mediasoup-client 3.18.8.

evgmel:
@ibc Thank you!

